PR #232, opened 2024-08-27
fix(p4_configure): resolve script execution errors and repair broken …

GrzesiekO:
```
---
title: Fix p4_configure.sh and crontab issues
issue_number: #218 
---

This commit fixes the errors that prevented the `p4_configure.sh` script from running correctly. It also addresses issues in the crontab configuration that caused scheduled tasks to fail.

No new features were introduced.

## Summary

### Changes

This commit includes the following changes:

1. Fixed syntax errors in the `p4_configure.sh` script that were causing it to fail during execution.
2. Corrected the crontab configuration to use the proper file paths and command syntax, ensuring that scheduled tasks run as expected.

### User experience

Before this change, users would encounter errors when attempting to run the `p4_configure.sh` script, and scheduled tasks would fail to execute. After this fix, the script runs successfully, and the crontab configuration works as intended, allowing scheduled tasks to complete as expected.

## Checklist

- [x] Changes are documented

<details>
No, this change does not introduce any breaking changes. It simply fixes existing issues to restore the expected functionality.
</details>

## Acknowledgment

### DIFF
--- a/assets/packer/perforce/helix-core/p4_configure.sh
+++ b/assets/packer/perforce/helix-core/p4_configure.sh
@@ -132,7 +132,7 @@ prepare_site_tags() {
 log_message "Starting the p4 configure script."
 
 # Check if the script received 7 arguments
-if [ "$#" < 6 || "$#" > 7 ]; then
+if [[ "$#" < 6 || "$#" > 7 ]]; then
     log_message "Incorrect usage. Expected 6 or 7 arguments, got $#."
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by jorisdon — APPROVED** [review]

**Comment by jorisdon** [comment]
@GrzesiekO We require signed commits. Can you set up a GPG key and amend a signature on this commit?

File: assets/packer/perforce/helix-core/p4_configure.sh [hunk 1]
```diff
@@ -132,7 +132,7 @@ prepare_site_tags() {
 log_message "Starting the p4 configure script."
 
 # Check if the script received 7 arguments
-if [ "$#" < 6 || "$#" > 7 ]; then
+if [[ "$#" < 6 || "$#" > 7 ]]; then
```
henrykie, line 135: I'm not convinced this is the right number of arguments. Can you check against the modules.

henrykie (approved):

PR #313, opened 2024-09-25
Added support for CASE_SENSITIVE setting

### LINKED ISSUE #302: Feature request: Ability to set case sensitivity in Helix Core
### Use case

Helix Core `p4d` has a list of startup flags that can be passed in to achieve different things (a full list can be found [here](https://www.perforce.com/manuals/p4sag/Content/P4SAG/appendix.p4d.html) ). 

For example, both Windows and UGS require the server to be run in case insensitive mode, which is the `-C1` startup flag.  

### Solution/User Experience

Add some ability to set the case sensitivity in the `helix-core` module on setup.
### Alternative solutions

_No response_

keith-miller:
resolves #302

**Issue number:**

### Changes

Added support for CASE_SENSITIVE setting

### User experience

## Checklist

* [ X] I have performed a self-review of this change
* [ X] Changes have been tested
* [ ] Changes are documented

<details>

</details>

## Acknowledgment

### DIFF
--- a/modules/jenkins/versions.tf
+++ b/modules/jenkins/versions.tf
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "5.66.0"
+      version = "5.68.0"
     }
     random = {
       source  = "hashicorp/random"
--- a/modules/perforce/helix-authentication-service/versions.tf
+++ b/modules/perforce/helix-authentication-service/versions.tf
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "5.66.0"
+      version = "5.68.0"
     }
     awscc = {
       source  = "hashicorp/awscc"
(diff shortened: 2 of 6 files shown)


keith-miller:
This is my first pass @henrykie which works, but I'm sure you will want to make the wording better :)

henrykie:
@GrzesiekO take a look...

henrykie:
@keith-miller this looks great to me. If it works for you then I'll let @GrzesiekO review - he may want to use this approach for now. We've been going back and forth on replacing this bash script entirely eventually, and as new requests for modifications come in we're leaning more and more that way.

We still need to set unicode mode. My understanding is you can do that against the server once it is already booted? We will still want to consider doing it by default.

keith-miller:
@henrykie that's correct. We were able to enable unicode support after the fact and are unblocked using this code.

GrzesiekO:
Seen the changes need to test it but seems like a quick fix following current pattern. Thank you @keith-miller!

GrzesiekO (approved):
All changes with current convention. Adds case sensitivity option for p4d. 
Approving.

henrykie (approved):

PR #383, opened 2024-11-05
chore: add external tls access to ALB + force cleanup resource flag

stephenlacy:
I noticed that the horde example stack doesn't have internet access through the ALB due to the security group rules.
Also added a enable_force_cleanup variable to cleanup s3 buckets.

### Changes
- Adds a flag to enable a clean stack deletion
- Enables public access to the horde server running on :443

## Checklist

* [ ] Changes are documented

## Acknowledgment

### DIFF
--- a/modules/unreal/horde/examples/complete/dns.tf
+++ b/modules/unreal/horde/examples/complete/dns.tf
@@ -46,6 +46,11 @@ resource "aws_acm_certificate" "unreal_engine_horde" {
   domain_name       = "horde.${data.aws_route53_zone.root.name}"
   validation_method = "DNS"
 
+  subject_alternative_names = [
+    "*.${data.aws_route53_zone.root.name}",
+    "*.horde.${data.aws_route53_zone.root.name}"
+  ]
+
   tags = {
     environment = "dev"
   }
@@ -63,6 +68,8 @@ resource "aws_route53_record" "unreal_engine_horde_cert" {
       record = dvo.resource_record_value
       type   = dvo.resource_record_type
     }
+    # skips the domain if it doesn't contain a wildcard
+    if length(regexall("\\*\\..+", dvo.domain_name)) > 0
   }
 
   allow_overwrite = true
--- a/modules/unreal/horde/sg.tf
+++ b/modules/unreal/horde/sg.tf
@@ -95,6 +95,18 @@ resource "aws_vpc_security_group_egress_rule" "unreal_horde_outbound_ipv6" {
   ip_protocol       = "-1" # semantically equivalent to all ports
 }
 
+
+# Inbound access to ALB from external traffic.
+resource "aws_vpc_security_group_ingress_rule" "unreal_horde_inbound_external_alb_traffic" {
+  count             = var.create_external_alb ? 1 : 0
+  security_group_id = aws_security_group.unreal_horde_external_alb_sg[0].id
+  description       = "Allow inbound web server traffic from Unreal Horde external ALB."
+  cidr_ipv4         = "0.0.0.0/0"
+  from_port         = var.external_api_port
+  to_port           = var.external_api_port
+  ip_protocol       = "tcp"
+}
+
 # Inbound access to Containers from External ALB on API port
 resource "aws_vpc_security_group_ingress_rule" "unreal_horde_inbound_external_alb_api" {
   count                        = var.create_external_alb ? 1 : 0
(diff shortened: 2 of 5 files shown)


modules/unreal/horde/sg.tf
```diff
@@ -95,6 +95,18 @@ resource "aws_vpc_security_group_egress_rule" "unreal_horde_outbound_ipv6" {
...
+resource "aws_vpc_security_group_ingress_rule" "unreal_horde_inbound_external_alb_traffic" {
+  count             = var.create_external_alb ? 1 : 0
+  security_group_id = aws_security_group.unreal_horde_external_alb_sg[0].id
+  description       = "Allow inbound web server traffic from Unreal Horde external ALB."
+  cidr_ipv4         = "0.0.0.0/0"
+  from_port         = var.external_api_port
+  to_port           = var.external_api_port
+  ip_protocol       = "tcp"
+}
+
```
henrykie, line 109: Hey! This is deliberately closed. We didn't want to have folks unknowning deploy the sample and open their ALB to the internet. One approach could be to comment it out? Another could be to provide a variable at the sample level that can be used to configure inbound access to the ALB. This is why we expose the SG ID as an output from the module.

modules/unreal/horde/examples/complete/dns.tf
```diff
@@ -46,6 +46,11 @@ resource "aws_acm_certificate" "unreal_engine_horde" {
   domain_name       = "horde.${data.aws_route53_zone.root.name}"
   validation_method = "DNS"
 
+  subject_alternative_names = [
+    "*.${data.aws_route53_zone.root.name}",
+    "*.horde.${data.aws_route53_zone.root.name}"
+  ]
```
henrykie, line 52: Why do we need to open up these certs for the sake of the example? Not opposed - just curious :)

modules/unreal/horde/variables.tf
```diff
@@ -74,6 +80,13 @@ variable "container_name" {
   nullable    = false
 }
 
+variable "external_api_port" {
+  type        = number
+  description = "The external port for the Unreal Horde web server."
+  default     = 443
+  nullable    = false
+}
+
```
henrykie, line 89: We need to make sure this propagates appropriately to the ALB listeners that are configured

henrykie (changes requested):
Couple of comments! Thanks for the PR. Would be happy to get these adjustments made.

modules/unreal/horde/examples/complete/dns.tf
stephenlacy: This is in part to combine with the samples/ stack of having multiple sub domains on the same root domain cert rather than creating a single cert for each horde deployment.

modules/unreal/horde/sg.tf
stephenlacy: Gotcha, I added OIDC to our deployment. Will add it as a variable.

gabebatista: I agree that having a variable (which defaults to no internet access) would be good solution. We can also simply call this out in the documentation since modifying a SG is a pretty simple task. Having the ingress rule there but commented out would make it even simpler.

modules/unreal/horde/variables.tf
```diff
@@ -38,7 +38,13 @@ variable "tags" {
...
   description = "Set this flag to enable ECS execute permissions on the Unreal Horde container and force new service deployments on Terraform apply."
-  default     = false
+  default     = true
+}
+
+variable "enable_force_cleanup" {
+  type        = bool
+  description = "Set this flag to true to enable force destroying of resources."
+  default     = true
 }
```
gabebatista, line 48: Would it make sense to have force cleanup tied to the debug flag? Are there any use cases where we would want one without the other?

modules/unreal/horde/sg.tf
henrykie: Resolving this - required fix is to remove this ingress rule from the module. We manage these at the sample level.

modules/unreal/horde/variables.tf
henrykie: Also, the actual service is not configured to run on 443 - we handle HTTPS -> HTTP termination at the ALB level. This is something I am open to parameterizing, but would require changes in multiple places.

henrykie:
@stephenlacy there were a couple of requests for changes. We'll close this PR (happy to review again) next week unless you've got updates. Thanks!

PR #386, opened 2024-11-08
fix(horde): add JwtIssuer to ensure container retains agents on restart

stephenlacy:
### Changes

In order to retain agents on ECS container restart the `JwtIssuer` must be set to a value that won't change. By default horde uses `Dns.GetHostName()` which changes on restart.

* [ ] Changes are documented

## Acknowledgment


pasotee:
I was just looking into this the other day. Thanks for this!

jorisdon (approved):

henrykie:
Thanks @stephenlacy !

PR #444, opened 2025-01-16
[DRAFT] Helix Swarm email configuration with Amazon SES

### LINKED ISSUE #375: Feature request: Ability to configure Swarm to send outgoing emails
### Use case

We would like to be able to set up Swarm to be able to send outgoing emails from Swarm notifications. The instructions on how to set up Swarm for mail can be found here:

https://www.perforce.com/manuals/swarm/Content/Swarm/admin.email.html

### Solution/User Experience

The Swarm Terraform module would set up Simple Email Service and then use it's configuration to set up Swarm email settings.

Found this SES TF module: https://github.com/cloudposse/terraform-aws-ses

### Alternative solutions

The Swarm Terraform module would allow users to pass in email settings if they wanted to use their own email setup.

henrykie:
**Issue number:**
closes #375 

### Changes

### User experience

## Checklist

* [ ] Changes are documented

<details>

</details>

## Acknowledgment


novekm:
Status as of 2:31am 1/16:
TASKS:
1. Modify bash script to conditionally remove the cache to use custom config.php file ✅
2. Dynamically create the config.php file. 
  a. Create core template of what is expected in the config.php file (`helix-swarm/main.tf:215`)
  b. Using variables (object) with conditionals, dynamically modify the default values of the template (`helix-swarm/main.tf:305-311`)
3. Upload the new config.php file to the Helix Swarm container 
  a. need to figure out optimal way to do this. Maybe TF local file generation -> upload file to S3 -> TF local-exec (run command to fetch this file using command in ECS)? Or maybe some other way from ECS itself to check for changes to S3 objects (using hash) and pull new version whenever hash changes
4. (stretch) Allow users to alternatively supply their own fully complete custom config.php file instead of using module input variables. This would be helpful for quick testing/troubleshooting purposes

novekm:
@henrykie question about this snippet in `modules/perforce/helix-swarm`:
```hcl
{
        name      = local.helix_swarm_data_volume_name
        image     = "bash"
        essential = false
        // Only run this command if enable_sso is set
        command = concat([], var.enable_sso ? [
          "sh",
          "-c",
          "echo \"/p4/a\\\t'sso' => 'enabled',\" > ${local.helix_swarm_data_path}/sso.sed && sed -i -f ${local.helix_swarm_data_path}/sso.sed ${local.helix_swarm_data_path}/config.php && rm -rf ${local.helix_swarm_data_path}/cache",
        # ] : []),
```
Questions:
1. How do you prevent a race condition where this command could run before the default `config.php` file is generated? There's a conditional that replaces a portion of this file depending on whether `var.enable_sso` is set or not. Is it assumed that the `config.php` file will always exist because it's already included in the image that comes from [perforce/helix-swarm](https://hub.docker.com/r/perforce/helix-swarm) or something?
2. With this snippet, would the command only run once? Is there a way we could re-trigger this command in an event-driven manner, such as an event being received in Amazon EventBridge (e.g. S3:PutObject operation from upload of a custom `config.php` file to an S3 bucket)?

henrykie:
> Maybe we can omit EventBridge+Lambda or similar and instead use though generally a last resort, use`local-exec`. When the custom `config.php` file is generated/update, we can maybe just use a `null_resource` with `local-exec` provisioner for AWS CLI with `aws ecs execute-command`
> 
> ex:
> 
> ```hcl
> resource "null_resource" "update_helix_swarm_config" {
>   # Will only trigger this resource to re-run if changes are made to the custom config.php file via the local_file resource
>   triggers = {
>     src_hash = local_file.custom_config_php.content_sha256
>   }
> 
>   # Fetch
>   provisioner "local-exec" {
>     command = <<-EOF
>      # 1. Copy custom config.php file from S3 bucket to ECS cluster within swarm data path
>       aws ecs execute-command --region ${data.aws_region.current} --cluster ${var.cluster_name} --task $TASK_ID (not sure how to get this yet...maybe environment variables within the ECS service?) --container-name $CONTAINER_NAME --command "aws s3 cp s3://${aws_s3_bucket.helix_swarm_config_bucket.id}/config.php ${local.helix_swarm_data_path}/config.php"
>       
>      # 2. Remove the cache so swarm loads the new config.php file
>       aws ecs execute-command --region ${data.aws_region.current} --cluster ${var.cluster_name} --task $TASK_ID  --container-name $CONTAINER_NAME --command "rm -rf ${local.helix_swarm_data_path}/cache"
> 
>     EOF 
>     
>   }
> 
>   # Only attempt to put the file when the S3 Bucket (and related resources) are created
>   depends_on = [
>     aws_s3_bucket.helix_swarm_config_bucket,
>     aws_s3_bucket_notification.helix_swarm_config_bucket,
>     aws_s3_bucket_policy.helix_swarm_config_bucket,
>     aws_s3_bucket_versioning.helix_swarm_config_bucket
>   ]
> }
> ```

If the task definition depends on S3 attribute path and version, then new versions of the TF resource will prompt new deployments of the task. Its even simpler!

novekm:
Status as of `1/16/25 11:20pm` 
### Status
- Successfully able to dynamically generate the custom `config.php` file based on input variables passed into the Helix-Swarm module.
- Successfully able to create the required SMTP (IAM) credentials for Amazon SES. Relevant values from SES (smpt host, username, password, etc.) are passed into the Helix-Swarm module at the sample level. If not using SES, users can pass in their relevant SMTP configurations (or otherwise) as desired, the module will function the same and generate the custom `config.php` file either way.
- Successfully able to upload the custom `config.php` file to the new `Helix Swarm Config` Amazon S3 Bucket.
- Subsequent programmatic changes to the custom `config.php` file triggers upload of a new version of the file to the Amazon S3 Bucket
- Necessary Environment Variables related to the custom config.php file are passed into the `bash` ECS Task Definition for later use (pull file from S3, etc.)

### Remaining Tasks
- Update the `bash` ECS Task Definition to run relevant commands to fetch the custom `config.php` file from the S3 Bucket, load it in to the Swarm data path to overwrite the previous version of this file, and remove the cache so Swarm will start using the new `config.php` file. Sub-tasks:
1. First will need to install the AWS CLI and then run the following commands (these commands may need to be modified):
```sh
# Install the AWS CLI
curl "https://awscli.amazonaws.com/awscli-exe-linux-aarch64.zip" -o "awscliv2.zip"
unzip awscliv2.zip
sudo ./aws/install
```
2. Copy file from S3 to Helix Swarm Data Directory
Ex. With Terraform Variables
```hcl
aws s3 cp s3://${aws_s3_bucket.helix_swarm_config_bucket.id}/config.php ${local.helix_swarm_data_path}/config.php
```
Ex. With Env Vars (created from Terraform Variables)
```hcl
aws s3 cp $HELIX_SWARM_CONFIG_S3_OBJECT_S3_URI $HELIX_SWARM_CONFIG_PHP_PATH
```
3. Delete the cache to reload swarm using the new config
Ex. With Terraform Variables
```hcl
rm -rf ${local.helix_swarm_data_path}/cache
```
Ex. With Env Vars (created from Terraform Variables)
```hcl
rm -rf $HELIX_SWARM_DATA_CACHE_PATH
```

4. Configure this task definition to be dependent on the S3 Object so the new versions of the task definition is created whenever new versions of the config.php file are created in S3. TBD on best way to do this.

henrykie:
@novekm I wanted to provide an update on where we are with this before today's meeting.

We are now able to do the following:

- Injection of custom `config.php` file into the Helix Swarm container prior to startup via S3 and sidecar.
- Creation of relevant SES resources that are dynamically added to `config.php`

However, this has unintended consequences:

- Because Helix Swarm detects the `config.php` prior to running `swarm-docker-setup.sh` it does not run a number of important configuration steps:
  - Install Helix Swarm extension on Helix Core server
  - Create Swarm p4 user
- The Apache errors (httpd not running) persist, even though the application is successfully running

So, what next?

After consulting folks at Perforce, I've established a couple of options.

1. Modify the image and entrypoint.

This would allow us to continue dynamically adding `config.php` prior to startup. We would then have to configure the Swarm container to install extensions (`swarm-extctl.sh`) ourselves. If complete customization at deployment is our goal this is the best path forward. From a support standpoint this could get fuzzy though - these scripts are complex, and we don't want to be in the business of updating Swarm business logic.

2. Provide a "basic" configuration by default, but persist the volume.

This approach is simple. It solves the problem where customer `config.php` is reset after the service restarts, but does not provide a mechanism to customize Swarm using IaC. Instead, we link the documentation and provide guidance on accessing the container using execute mode to modify the `config.php` file and remove the cache.

What about SES authentication using IAM?

I also consulted Perforce on this topic. They indicated that the `connection_class` information is all handled via [Laminas](https://docs.laminas.dev/laminas-mail/) and there is already an [SES related issue there.](https://github.com/laminas/laminas-mail/issues/118)

PR #458, opened 2025-01-30
Add iscsci support

adizalmanovich1:


modules/perforce/fsxn_resources/resources.tf
```diff
@@ -0,0 +1,149 @@
...
+    size      = var.depot_volume_size
+    size_unit = "gb"
+  }
+  nas = {
+    junction_path = "/hxdepots"
+  }
+  depends_on = [aws_lambda_function.lambda_function]
+}
+
+resource "netapp-ontap_lun" "depots_volume_lun" {
```
gabebatista, line 125: I've noticed that the resource names in this Terraform configuration tend to use a mix of hyphens and underscores. Is there a particular naming convention or pattern being followed here?

modules/perforce/fsxn_resources/link.tf
gabebatista, line 1: Could you provide some context on what the lambda function does here? I'm assuming that it does some configuration to connect perforce with FSxN?

modules/perforce/fsxn_resources/resources.tf
adizalmanovich1: can you explain?

modules/perforce/fsxn_resources/link.tf
adizalmanovich1: The Lambda is created to communicate directly with ONTAP to create NetApp resources, such as volumes, luns, and igroups.

modules/perforce/fsxn_resources/resources.tf
gabebatista: We have been using underscores as a naming convention for Terraform resource (e.g. `netapp_ontap_lun`) but i noticed in the code you are using a mix of hyphens and underscored (e.g `netapp-ontap_lun`). I was just curious about your thought process here. I hope my questions don't come off as nit-picky, I'm just curious whether that is an established naming convention within Terraform. If not, I may ask you to change this to keep the naming convention standard across the toolkit.

LirazRom10: this is Netapp TF resource naming convention
https://registry.terraform.io/providers/NetApp/netapp-ontap/latest/docs/data-sources/lun

we can ask them about it but change will break backward compatability

modules/perforce/fsxn_resources/link.tf
LirazRom10: the Netapp TF dont have a resource to deploy a lambda, actually the lamnda addition in their code came from our request to support our link - which is basically the lambda that is being deployed in this solution. we have developed a lambda image that is basically a proxy - it live inside the customer vpc and can be invoked using aws sdk/cli. it support both https calls and ssh sessions.

modules/perforce/fsxn_resources/resources.tf
gabebatista: If changes will break backwards compatibility then we can just leave it as is, no need to complicate things 😄

modules/perforce/examples/fsxn/main.tf
```diff
@@ -0,0 +1,63 @@
+##########################################
+# FSx ONTAP File System
+##########################################
+
+resource "aws_fsx_ontap_file_system" "helix_core_fs" {
+  storage_capacity    = 1024
+  subnet_ids          = [var.instance_subnet_id]
+  preferred_subnet_id = var.instance_subnet_id
+  deployment_type     = "SINGLE_AZ_1"
```
h0lmstr0m, line 9: Question, there is discussion right now with Riot and Epic that may result in a recommendation of Multi-AZ FSxN and Primary/Secondary Helix Core servers as the primary recommendation for running Perforce in AWS on FSxN. Should that be defined here, or is that better for "guidance" in the documentation?

henrykie: We will do this. This example showcases a single Helix Core server deployment. We do not currently have support to configure Helix Core servers as Edge / Replica / Standby, although @GrzesiekO is working on that currently.

henrykie (changes requested):
Couple minor things we need to adjust:
- Rebase main to pull in changes to packer base AMI version
- Potentially update example to create VPC and handle DNS to follow existing patterns

I can add these changes.

modules/perforce/examples/fsxn/main.tf
h0lmstr0m, line 1: I may just be missing it, but is autogrow for the Helix Core luns configured as part of the TF here? Same for the FS itself? I believe it's an assumption that the volumes will be elastic like EBS.

assets/packer/perforce/helix-core/perforce_arm64.pkr.hcl
```diff
@@ -50,7 +50,7 @@ source "amazon-ebs" "al2023" {
 
   source_ami_filter {
     filters = {
-      name                = "al2023-ami-2023.5.*"
```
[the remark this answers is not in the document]
LirazRom10: ok, merged it to the fork

modules/perforce/examples/fsxn/variables.tf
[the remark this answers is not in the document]
LirazRom10: adding all

LirazRom10:
@henrykie i have added the missing items to the sample 
now it is like the regular complete
in addition also rebased so the ami changes there

henrykie:
Still working on refactoring. Should have updates soon. Sorry for the delay!

PR #585, opened 2025-04-20
refactor: Perforce modules consolidated to simplify shared resource creation

### LINKED ISSUE #565: Feature request: Consolidate Perforce modules
### Use case

This would entail consolidating the Helix Core, Helix Swarm, and Helix Authentication Service into a single Perforce Module. This would also use the updated branding from Perforce (P4, P4 Code Review, P4 Auth). 

### Solution/User Experience

Users could use a single module declaration and conditionally use variables to deploy each service:

```hcl

# This is a template file for a basic deployment.
# Modify the parameters below with your desired values.

module "perforce" {
  source = "../../"
  # source = "aws-games/perforce/aws"
  # version = "v1.0.0"

  # SHARED
  fully_qualified_domain_name = "novekm.people.aws.dev"

  create_vpc           = true
  vpc_cidr_block       = "10.0.0.0/16"
  public_subnet_cidrs  = ["10.0.1.0/24", "10.0.2.0/24"]
  private_subnet_cidrs = ["10.0.3.0/24", "10.0.4.0/24"]

  # HELIX CORE
  helix_core_config = {
    resource_prefix = "helix-core"
    # helix_authentication_service_url = optional(string, null)

    # Compute
    lookup_existing_ami      = false
    enable_auto_ami_creation = true
    ami_prefix               = "p4_al2023"

    instance_type         = "c6g.large"
    instance_architecture = "arm64"
    p4_server_type        = "p4d_commit" # required
    unicode               = false
    selinux               = false
    case_sensitive        = true
    plaintext             = false

    # Storage
    storage_type = "EBS"

    depot_volume_size    = 128
    metadata_volume_size = 32
    logs_volume_size     = 32

    # Networking
    lookup_existing_instance_subnet_id = false
    # instance_subnet_id       = optional(string, null)
    # existing_security_groups = optional(list(string), [])
    internal = false

    # Security & Auth
    # super_user_username_secret_arn = optional(string, null)
    # super_user_password_secret_arn = optional(string, null)

    # existing_helix_core_role       = optional(string, null)
    create_helix_core_default_role = true

  }

  # HELIX SWARM
  # helix_swarm config = {
  #   server_type = ""

  # }

  # # HELIX AUTH SERVICE
  # helix_auth_service_config = {

  # }
}


```

### Alternative solutions

Leave as is

### LINKED ISSUE #571: Bug: Perforce Helix Core Destroy Failure
### Expected Behaviour

Successful termination of resources

### Current Behaviour

Must do a second `terraform destroy` due to a dependency on the perforce-web-services ALB Listener and target group:

```
│ Error: deleting ELBv2 Listener (arn:aws:elasticloadbalancing:us-east-1:xx:listener/app/perforce-web-services/xx/xx): operation error Elastic Load Balancing v2: DeleteListener, https response error StatusCode: 400, RequestID: xx, ResourceInUse: Listener port '443' is in use by registered target 'arn:aws:elasticloadbalancing:us-east-1:xx:loadbalancer/app/perforce-web-services/xx' and cannot be removed.
```

### Possible Solution

Can potentially add lifecycle rules and explicit dependencies with `depends_on` with similar configuration as this:
```
resource "aws_lb_listener" "front_end" {
  # ... other configuration ...

  lifecycle {
    create_before_destroy = true
  }

  depends_on = [aws_lb_target_group.example]
}

resource "aws_lb_target_group" "example" {
  # ... other configuration ...

  lifecycle {
    create_before_destroy = true
  }
}
```

### Steps to Reproduce

This intermittently happens during a run of `terraform destroy`

latest

### LINKED ISSUE #574: Feature request: Automate deletion of non-empty ALB Access Log buckets when `debug` is `true`
### Use case

Currently, when destroying a module that provisions an ALB access log bucket during debugging the `terraform destroy` fails because the bucket is not empty. This slows down debugging.

### Solution/User Experience

Add `force_destroy = var.debug` for module ALB access log buckets.

### Alternative solutions

_No response_

### LINKED ISSUE #575: Feature request: Perforce module ALB access log bucket force delete on debug
### Use case

See #574 

### Solution/User Experience

See #574 

### Alternative solutions

_No response_

novekm:
**Issue number:** 

closes #565 
closes #571 
closes #574 
closes #575

### Changes

This PR consolidates the three existing Perforce modules (**Helix Core**, **Helix Swarm**, and **Helix Authentication Service**) into a single parent module (`terraform-aws-perforce`) with the three original modules becoming child modules of the parent module.  The child modules and related files have also all been updated to reflect the new Perforce branding, with products such as **P4 Server** (formerly Helix Core), **P4 Auth** (formerly Helix Authentication Service) and **P4 Code Review** (formerly Helix Swarm).

### User experience

Before this PR users had to explicitly create module declarations for each individual module, as well as initialize and manage the versions of all thee modules individually. With these changes, they can use a single module (terraform-perforce-aws) and use the optional variables such as `p4_server_config`, `p4_auth_config`, and `p4_code_review_config` to conditionally deploy the related resources.

## Checklist

* [ ] Changes are documented 

<details>
Yes, this will change the way users can leverage the modules, and a vast number of resource names have been changed.
</details>

## Acknowledgment


henrykie:
@novekm I'm going to convert this to a draft until we've got time to clean up the commit history

modules/perforce/sg.tf
```diff
@@ -0,0 +1,256 @@
...
+  from_port         = 443
+  to_port           = 443
+  cidr_ipv4         = var.public_subnets_cidrs[0] != null ? var.public_subnets_cidrs[0] : var.private_subnets_cidrs[0]
+
+  tags = merge(var.tags,
+    {
+      Name = "${var.project_prefix}-allow-nlb-subnet-1"
+    }
+  )
+}
```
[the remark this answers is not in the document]
henrykie: basically, we shouldn't be using CIDRs. THis allows ingress access to the web services ALB from the entire VPC... We just want inbound from the P4 server (using SGs) so it can access auth, and from the NLB for external access.

modules/perforce/sg.tf
```diff
@@ -0,0 +1,256 @@
...
+# # Inbound to the P4 Server Security Group from the P4 Code Review Security Group
+# resource "aws_vpc_security_group_ingress_rule" "p4_server_inbound_p4_code_review" {
+#   count                        = var.create_default_sgs != false ? 1 : 0
+#   security_group_id            = module.p4_server[0].security_group_id
+#   description                  = "P4 Code Review inbound to P4 Server"
+#   ip_protocol                  = "TCP"
+#   from_port                    = 1666
+#   to_port                      = 1666
+#   referenced_security_group_id = module.p4_code_review[0].service_security_group_id
+# }
```
henrykie, line 256: Cleanup?

modules/perforce/variables.tf
```diff
@@ -0,0 +1,600 @@
...
+  type        = string
+  description = "The current environment (e.g. dev, prod, etc.)"
+  default     = "dev"
+}
+
+variable "enable_shared_alb_access_logs" {
+  type        = bool
+  description = "Whether to enable access logs for the shared ALB."
+  default     = true
+}
```
henrykie, line 31: Is this duplicated with the variable below: `enable_shared_lb_access_logs (56)`

modules/perforce/variables.tf
```diff
@@ -0,0 +1,600 @@
...
+variable "shared_nlb_access_logs_prefix" {
+  type        = string
+  description = "Log prefix for shared NLB access logs."
+  default     = null
+}
+variable "shared_alb_access_logs_prefix" {
+  type        = string
+  description = "Log prefix for shared ALB access logs."
+  default     = null
+}
```
henrykie, line 77: I like a consolidated Access Logs bucket. Did you validate that this is how the parent module works?

modules/perforce/variables.tf
```diff
@@ -0,0 +1,600 @@
...
+
+variable "s3_enable_force_destroy" {
+  type        = bool
+  description = "Enables force destroy for the S3 bucket for both the shared NLB and shared ALB access log storage. Defaults to true."
+  default     = true
+}
```
henrykie, line 83: I think this should just depend on `debug` instead of a new variable. What are your thoughts?

modules/perforce/variables.tf
```diff
@@ -0,0 +1,600 @@
...
+
+  validation {
+    condition     = length(var.p4_server_config.name) > 1 && length(var.p4_server_config.name) <= 50
+    error_message = "The defined 'name' has too many characters (${length(var.p4_server_config.name)}). This can cause deployment failures for AWS resources with smaller character limits. Please reduce the character count and try again."
+  }
+
+  validation {
+    condition     = var.p4_server_config.instance_architecture == "arm64" || var.p4_server_config.instance_architecture == "x86_64"
+    error_message = "The p4_server_config.instance_architecture variable must be either 'arm64' or 'x86_64'."
+  }
```
henrykie, line 306: Do we need to do this level of validation? Or would the `plan` fail if we passed something different here due to the validation at the submodule level?

modules/perforce/variables.tf
```diff
@@ -0,0 +1,600 @@
...
+########################################
+# P4 Server
+########################################
+variable "p4_server_config" {
+  type = object({
+    # General
+    name                        = optional(string, "p4-server")
+    project_prefix              = optional(string, "cgd")
+    environment                 = optional(string, "dev")
+    service_url                 = optional(string, null)
```
henrykie, line 193: `authentication_service_url`

modules/perforce/variables.tf
```diff
@@ -0,0 +1,600 @@
...
+    fully_qualified_domain_name = string
+
+    # Compute
+    lookup_existing_ami      = optional(bool, true)
+    enable_auto_ami_creation = optional(bool, false)
+    ami_prefix               = optional(string, "p4_al2023")
+
+    instance_type         = optional(string, "c6g.large")
+    instance_architecture = optional(string, "arm64")
```
henrykie, line 202: we're mixing our defaults between this config block and the sub-module variables (x86, c6in). We should make sure they're in sync.

henrykie (changes requested):
Awesome job. This is a lot of very nitty-gritty work in the Perforce modules, so I hope you came out the other side with a better understanding of how all these tools work together. You'll find my comments / requests for changes inline.

One overall note:

References to `P4Auth` versus `P4 Authentication Service`; I think if we are going to do all this name changing it makes sense to use the same name as the partner branding 😿 But I also don't want to make you go back through and find all of them. I'll leave it up to you...

modules/perforce/examples/create-resources-complete/README.md
```diff
@@ -0,0 +1,71 @@
+# Create Resources Complete
+This example demonstrates how to deploy **[P4 (formerly P4 Server)](https://www.perforce.com/products/helix-core)**, **[P4 Code Review (formerly P4 Code Review)](https://www.perforce.com/products/helix-swarm)**, and **[P4Auth (formerly P4Auth)](https://help.perforce.com/helix-core/integrations-plugins/helix-auth-svc/current/Content/HAS/overview-of-has.html)**.
+
+
+## Architecture
+
+![perforce-complete-arch](../../assets/media/diagrams/perforce-arch-cdg-toolkit-terraform-aws-perforce-full-arch-route53-dns.png)
+
```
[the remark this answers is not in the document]
gabebatista: If I remember correctly, READMEs on github do support markdown so the question is what special syntax are we using for mkdocs that falls outside of the "vanilla" markdown Github can render.

modules/perforce/examples/create-resources-complete/README.md
```diff
@@ -0,0 +1,71 @@
+# Create Resources Complete
+This example demonstrates how to deploy **[P4 (formerly P4 Server)](https://www.perforce.com/products/helix-core)**, **[P4 Code Review (formerly P4 Code Review)](https://www.perforce.com/products/helix-swarm)**, and **[P4Auth (formerly P4Auth)](https://help.perforce.com/helix-core/integrations-plugins/helix-auth-svc/current/Content/HAS/overview-of-has.html)**.
```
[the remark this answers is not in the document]
novekm: whoops, good catch - will update :)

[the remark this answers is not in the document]
novekm: Hmm, so I was confused on that - I was talking to someone from Perforce (they emailed me once I downloaded the CFN template to test that deployment). I asked about the new naming and they told me it's P4, P4Auth, and P4 Code Review. Happy to change, but also not sure if I need to fix the P4 Server/p4-server stuff too...

modules/perforce/examples/create-resources-complete/README.md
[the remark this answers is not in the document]
novekm: I'm open to it, but sometimes there are cases where you may need to add more relevant info and it's more readily accessible within the example itself. Here's [one example in another project](https://github.com/aws-ia/terraform-aws-iam-identity-center/tree/main/examples/existing-users-groups-create-apps) highlighting a specific error I wanted to call out that could arise if not using the module correctly. It's mentioned in the root README too, however that is a pretty long file and it can cause a substantial issue that's not immediately clear from the TF logs how to resolve. I feel like there may be situations like this that can arise, especially the more modular a module is.

Also, I think it may be helpful to have an arch diagram per example (within reason) to make it easier to visualize what will be deployed, though not necessary

modules/perforce/examples/create-resources-complete/dns.tf
```diff
@@ -0,0 +1,87 @@
...
+  records = [module.terraform-aws-perforce.p4_server_eip_public_ip]
+}
+
+
+##########################################
+# P4 Code Review Certificate Management
+##########################################
+resource "aws_acm_certificate" "perforce" {
+  domain_name               = local.acm_domain_name
+  subject_alternative_names = [local.acm_subject_alternative_names]
```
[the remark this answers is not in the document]
novekm: Gotcha, I'll remove `subject_alternative_names` and try another test deployment. Also, I'll update the value for `domain_name` to be:
```hcl
domain_name = "${local.perforce_subdomain}.${local.fully_qualified_domain_name}"
```
`locals.tf` currently has the following locals available related to subdomains):
```hcl
  # Subdomains
  perforce_subdomain       = "perforce"
  p4_auth_subdomain        = "auth"
  p4_code_review_subdomain = "review"
```

Mainly just wanted to avoid potential user error due to simple typos since these subdomains are used pretty heavily throughout the example. So perhaps we can just reuse these here too?

modules/perforce/examples/create-resources-complete/locals.tf
```diff
@@ -0,0 +1,43 @@
...
+
+locals {
+  project_prefix = "cgd"
+  azs            = slice(data.aws_availability_zones.available.names, 0, 2)
+
+  # Route53 - only set this if you have an existing Route53 Public Hosted Zone
+  route53_root_domain_name = "<replace with the name of your Route53 Public Hosted Zone>" # Ensure this a Route53 Public Hosted Zone with this name exists in your AWS Account
+
+  # Domain name
+  fully_qualified_domain_name = "<replace with the name of your domain>" # Ensure this domain already exists, either in Route53 with Public Hosted Zone, or in a 3rd party DNS provider.
```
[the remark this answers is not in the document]
novekm: So I removed the `variables.tf` file and the `root_domain_name` variable at the example level. As it was a required variable (no default value), and the example isn't technically a module (at least not one where you can add a module declaration for), it would force users to enter this value during every run of `terraform apply`. This is because you can't set the value of a Terraform variable within the same context of the module that is using it besides using a default or using a `.tfvars` file, or when using a module declaration with the `"module" block. This would also mean we'd have to make this file to set up the automated testing or the pipeline would hang when it asks for the var.

In my testing I goofed a few times with typos with the `root_domain_name`. I wanted to avoid potential user error around this (since it can cause deletions/recreations instead of a simple update in place) so thought it best to let this be a static value within a `locals` block. None of the other modules use a `.tfvars` file and instead use locals mainly, make users type these values in during `terraform apply` or manually create their own `.tfvars` file outside of the example, so I wanted to follow the same for the Perforce module. 

On the example`route53_domain_name` and `fully_qualified_domain_name` locals - I agree the `route_53_domain_name` local can be removed. Initially I was conditionally creating route53 resources at the example level depending on if this value was null or not, but must have removed this logic a while ago and forgot about it. However, I believe we should keep the `fully_qualified_domain_name` local for the reasons mentioned above

modules/perforce/examples/create-resources-complete/locals.tf
```diff
@@ -0,0 +1,43 @@
...
+  p4_auth_fully_qualified_domain_name = "${local.p4_auth_subdomain}.${local.perforce_subdomain}.${local.fully_qualified_domain_name}"
+
+  # P4 Code Review
+  p4_code_review_fully_qualified_domain_name = "${local.p4_code_review_subdomain}.${local.perforce_subdomain}.${local.fully_qualified_domain_name}"
+
+
+  # Subdomains
+  perforce_subdomain       = "perforce"
+  p4_auth_subdomain        = "auth"
+  p4_code_review_subdomain = "review"
```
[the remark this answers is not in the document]
novekm: yeah it might be - I just ran into a ton of small little issues because of typos between the modules. Since they're used in more than one place I think it's worth keeping them

modules/perforce/variables.tf
novekm: I think there may be a use case where someone wants to enable force destroy but may not want debug enabled. Also, I think from a user standpoint, debug being enabled would mean that logs should be retained to investigate? But I guess we could just have something like `force_destroy = var.debug == false ? false : true` to make it simpler. I just think it's a bit more clear to be explicit since it could accidentally delete logs that they needed to retain for some reason

modules/perforce/variables.tf
novekm: It's used in the `local-exec` and `local-file` so I think we need to keep it. Plan wouldn't fail for those

modules/perforce/variables.tf
novekm: Can we keep this for now? I'd have to change this in a few places (`p4_configure.sh, `local-file.tf`) and a quite a few other file. I did a find across the repo and it's used 24 times, but some of them are part of a longer string like `HELIX_AUTH_SERVICE_URL` so it would probably cause breaking changes unless I went file by file. But I agree that it likely should be more clear. My for is for `auth_service_url` if we're keeping the `p4_auth` stuff, or `authentication_service_url` as you recommended if we change all of the `p4_auth` stuff to be `p4_authentication_service` or if you wanna just push a commit to this branch feel free 🙂

modules/perforce/variables.tf
novekm: Ah gotcha, updated all to the graviton options

gabebatista (commented):
I think Henry hit on all the important pieces already, I just left some minor comments. Fantastic work and done in record time!

modules/perforce/variables.tf
novekm: nvm, I fixed it - used `auth_service_url` to follow current naming convention of `p4_auth` used elsewhere in the module

modules/perforce/examples/create-resources-complete/README.md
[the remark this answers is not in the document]
novekm: Hmm yeah that could work. I mainly did it for the future state of moving this (and the other modules) to its own standalone GitHub repo

modules/perforce/examples/create-resources-complete/main.tf
```diff
@@ -0,0 +1,63 @@
...
+    deregistration_delay        = 0
+  }
+
+
+  # - P4 Code Review Configuration -
+  p4_code_review_config = {
+    name                        = "p4-code-review"
+    fully_qualified_domain_name = local.p4_code_review_fully_qualified_domain_name
+    existing_security_groups    = [aws_security_group.allow_my_ip.id]
+    debug                       = true # optional to use for debugging. Default is false if omitted
```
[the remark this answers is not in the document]
novekm: I think we should keep it in the folder itself since the future state is to move it to it's own GitHub repo to be a standalone module. What do you think?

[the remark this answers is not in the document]
novekm: I think there might be - for example, if one service is working but the other isn't, you may just want to enable debug for just that service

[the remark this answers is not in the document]
novekm: Also, can't reply to the below comment so will add the reply here: 
> I like the idea of the NLB expecting a 301 and routing users to swarm, seems like a better user experience to me also.

I think we should keep the 200. When I had the 301 I was having issues with the health checks failing for the ALB target group. Also, in the off chance that someone just wanted to deploy P4Auth and P4 Server (and not P4 Code Review/Swarm), the 301 would route to nothing by default which I think may cause an issue. I don't think a default rule can point to non-reachable target, or if it can, I think the health checks may fail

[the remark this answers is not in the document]
henrykie: w.r.t. the `debug` conversation - I think it makes sense to provide a single `debug` per module. And if you enable `debug` at the `perforce` root level then it should propagate to all sub-modules. But not a single `debug` across all modules being deployed (including Horde / TC).

Remember the tenant that examples should require minimal external configuration. In the case of the Perforce example it should be **just** the `root_domain_name` I think.

modules/perforce/examples/create-resources-complete/locals.tf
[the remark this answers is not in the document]
henrykie: Wait, what? We've been using variables all over the place for examples:
- [TeamCity example](https://github.com/aws-games/cloud-game-development-toolkit/blob/cdc4dd5236d9cbe68dd1386857d9f53ea2babf87/modules/teamcity/examples/simple/dns.tf#L1-L4)
- [Horde example](https://github.com/aws-games/cloud-game-development-toolkit/blob/main/modules/unreal/horde/examples/complete/variables.tf)

The whole point of having a `root_domain_name` variable at the example level is _because_ we can't infer this. The end user has to provision a hosted zone before deploying the example. That is a requirement.

I am not following why it was removed. The objective of the `variables.tf` is to provide a single point of entry for configurables for the example.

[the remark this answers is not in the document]
novekm: Sorry let me clarify - with the above there isn't a way to supply this value besides during runtime (`terraform apply)`. For example, for the Jenkins module, there are two variables at the example level:
```hcl
variable "fully_qualified_domain_name" {
  type        = string
  description = "A fully qualified domain name (FQDN) to be used for jenkins. A record will be created on the hosted zone with the following patterns 'jenkins.<your_fqdn>'"
}

variable "jenkins_agent_secret_arns" {
  type        = list(string)
  description = "A list of secretmanager ARNs (wildcards allowed) that contain any secrets which need to be accessed by the Jenkins service."
  default     = []
}
```
For a user deploying this example, how would they currently be able to supply these variables? With the above, they would have to supply a value for `fully_qualified_domain_name` during every run of `terraform apply`, or they would have to create a `.tfvars` file that has this value. Then during apply, Terraform would load this file and use the variable defined here instead of prompting the user to enter it every time the run apply. The `jenkins_agent_secret_arns` has a default, so it wouldn't prompt the user to enter this during apply. This also means there is no way for them to supply the value unless they do one of the following:
- add their own default value to the variable
- make a `.tfvars` file and store the value there
- pass in the values inline when running `terraform apply` such as `terraform apply -var "jenkins_agent_secret_arns = ["arn1", "arn2"]"`

For the Perforce module, the same is the case - if you create a `variables.tf` file a user will have to enter this value continuously:
```hcl
❯ tfp
var.root_domain_name
  The root domain name

  Enter a value: 

```
or encounter an error:
```
│ Error: No value for required variable
│ 
│   on variables.tf line 1:
│    1: variable "root_domain_name" {
│ 
│ The root module input variable "root_domain_name" is not set, and has no default value.
│ Use a -var or -var-file command line argument to provide a value for this variable.
```

You can't define the variable outside of a module declaration unless using one of the other options mentioned above. For example:
```hcl
root_domain_name = "example.com" # <-- this is not valid

module "terraform-aws-perforce" {
  source = "../../"
  # ...additional configuration
}
```
The above will not work, throwing an error `Unexpected attributed: an attribute namee "root_domain"name" is not expected here.`

I don't think we need to create example level variables, especially those that are required. Instead, we should supply values for existing variables that are present within the terraform modules themselves, in the module declaration:

```hcl
module "terraform-aws-perforce" {
  source = "../../"

  # - Shared -
  project_prefix       = local.project_prefix
  vpc_id               = aws_vpc.perforce_vpc.id
  public_subnets       = aws_subnet.public_subnets[*].id
  public_subnets_cidrs = local.public_subnet_cidrs
  private_subnets      = aws_subnet.private_subnets[*].id
}
# etc
```

For example, see [the examples for `terraform-aws-vpc` module](https://github.com/terraform-aws-modules/terraform-aws-vpc/tree/master/examples). They have all have a `variables.tf` file, but none of them are actually even used - the files are all empty and instead they just use locals. This is likely for the reasons mentioned above. Users typically would create their own variables if desired, or they can just use locals. I don't think we need to be opinionated on the variables they should create at the example level, just the ones at the module level. I propose we just standardize with using `locals` at the example level.

modules/perforce/examples/create-resources-complete/main.tf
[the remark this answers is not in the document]
novekm: Ah yeah, I forgot about the listeners in the submodules. I'm happy to revert if needed

modules/perforce/examples/create-resources-complete/locals.tf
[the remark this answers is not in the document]
henrykie: Jenkins is the wrong example to leverage here. I am not suggesting that the Perforce root module has a variable called `root_domain_name` - the Perforce root module is not responsible for provisioning DNS related resources beyond the private hosted zone, which you capture with a `var.private_hosted_zone_name` and a `var.create_private_hosted_zone`.

The objective with an `example` is that it is deployable without modification. Using `locals` requires a user to edit the file to input their own values - in my opinion that is no more streamlined than having to define a `.tvars` or explicitly set variables on the command line. The `create-resources-complete` example _should_ prompt the user to input a `root_domain_name` on `terraform apply` since this example is creating public DNS resources in that hosted zone.

modules/perforce/examples/create-resources-complete/dns.tf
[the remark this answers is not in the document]
novekm: Updated

modules/perforce/examples/create-resources-complete/locals.tf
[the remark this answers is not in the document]
novekm: Thanks for the feedback all! Updated to add back in the variable at the example level and named it `route53_public_hosted_zone_name` for clairty

modules/perforce/main.tf
```diff
@@ -0,0 +1,203 @@
...
+#################################################
+module "p4_server" {
+  source = "./modules/p4-server"
+  count  = var.p4_server_config != null ? 1 : 0
+
+  # General
+  name                        = var.p4_server_config.name
+  project_prefix              = var.p4_server_config.project_prefix
+  environment                 = var.p4_server_config.environment
+  auth_service_url            = "https://${var.p4_server_config.fully_qualified_domain_name}"
```
henrykie, line 12: This is incorrect. `var.p4_server_config.fully_qualified_domain_name` would be `perforce.example.com`, correct? But this is the URL for the auth service. Should be:

`auth_service_url = var.p4_server_config.auth_service_url` (which includes `https://` since we cannot confirm what protocol the p4 auth service will be available on). however, since we are internally creating the private hosted zone there is a conditional where it would actually point at the internal record for the auth service (if its defined).

modules/perforce/main.tf
```diff
@@ -0,0 +1,203 @@
...
+  fully_qualified_domain_name = var.p4_server_config.fully_qualified_domain_name
+
+
+  # Compute
+  lookup_existing_ami      = var.p4_server_config.lookup_existing_ami
+  enable_auto_ami_creation = var.p4_server_config.enable_auto_ami_creation
```
henrykie, line 18: Same as before. Do we need both?

modules/perforce/main.tf
```diff
@@ -0,0 +1,203 @@
...
+  project_prefix                  = var.p4_auth_config.project_prefix
+  environment                     = var.p4_auth_config.environment
+  enable_web_based_administration = var.p4_auth_config.enable_web_based_administration
+  debug                           = var.p4_auth_config.debug
+  fully_qualified_domain_name     = var.p4_auth_config.fully_qualified_domain_name
+
+  # Compute
+  cluster_name = local.create_shared_ecs_cluster ? aws_ecs_cluster.perforce_web_services_cluster[0].name : var.existing_ecs_cluster_name
+
+  # cluster_name            = aws_ecs_cluster.perforce_web_services_cluster[0].name
```
henrykie, line 65: nit

modules/perforce/modules/p4-server/variables.tf
```diff
@@ -25,93 +24,94 @@ variable "environment" {
...
-  description = "A list of existing security group IDs to attach to the Helix Core load balancer."
-  default     = []
+variable "p4_server_type" {
+  type        = string
+  description = "The Perforce P4 Server type."
+  validation {
+    condition     = contains(["p4d_commit", "p4d_replica"], var.p4_server_type)
+    error_message = "${var.p4_server_type} is not one of p4d_commit or p4d_replica."
+  }
 }
```
henrykie, line 85: @GrzesiekO @jorisdon are we comfortable defaulting P4 server to Graviton?

modules/perforce/examples/create-resources-complete/main.tf
```diff
@@ -0,0 +1,64 @@
...
+
+  # - P4 Server Configuration -
+  p4_server_config = {
+    # General
+    name                        = "p4-server"
+    fully_qualified_domain_name = local.p4_server_fully_qualified_domain_name
+
+    # Compute
+    lookup_existing_ami      = false
+    enable_auto_ami_creation = true
```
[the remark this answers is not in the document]
novekm: I agree - I think it's worth as a team creating a laundry list of the possible customization options, creating examples for the one's we feel are most likely (with related tf tests), and variable validations for the rest (in addition to the common deployment options we expect)

modules/perforce/lb.tf
```diff
@@ -0,0 +1,339 @@
...
+##########################################
+# Used to set dependency on ALB from parent module, since depends_on won't work upstream
+# This triggers during the first apply, or if the ALB ARN changes to a different value, such as null
+resource "null_resource" "parent_module_certificate" {
+  # count = var.create_application_load_balancer
+  triggers = {
+    certificate_arn = var.certificate_arn
+  }
+
+}
```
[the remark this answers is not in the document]
novekm: This currently can likely be removed. I created this as an attempt to solve a race condition where the listener could attempt to use a certificate before validation finished causing this error:
```hcl
│ Error: modifying ELBv2 Listener (arn:aws:elasticloadbalancing:us-east-1:xx:listener/app/cgd-perforce-shared-alb/xx/xx): operation error Elastic Load Balancing v2: ModifyListener, https response error StatusCode: 400, RequestID: xx-xx-xx-xx-xx, api error UnsupportedCertificate: The certificate 'arn:aws:acm:us-east-1:xx:certificate/xx-xx-xx-xx-xx' must have a fully-qualified domain name, a supported signature, and a supported key size.
│ 
│   with module.terraform-aws-perforce.aws_lb_listener.perforce_web_services[0],
│   on ../../lb.tf line 161, in resource "aws_lb_listener" "perforce_web_services":
│  161: resource "aws_lb_listener" "perforce_web_services" {
```
 This doesn't occur during a normal deployment, since the way the dependency tree is set and how long the dependent resources take to provision, the certificate validation is complete before the listener tries to use the certificate. However, during an update to the certificate, this triggers a forced replacement as the cert needs to be re-generated and re-validated. In this case since it's only a few resources that change, the above error happens as the listener tries to use the cert before validation finishes.
 
I attempted to use this to try to force a dependency between the certificate (created at example level) and listener (created within the root perforce module) to ensure the listener would wait to be created/updated until the certificate arn changes. In my testing I switched this to be the certificate validation id in hopes that the ID would only be generated ***after*** the validation was complete, but unfortunately the id is generated during apply (same as the ARN of the certificate itself) so this functionality doesn't work. It doesn't hurt though.

I then attempted to try to have the entire creation of the perforce module resources depend on the successful creation of the cert validation, however this was triggering a forced replacement for multiple resources within the perforce module (though no changes had been made) because the changes were `known after apply` and most of the underlying resources don't support update in place.

**To ultimately solve for this I see two options**:
1. Move the creation creation of the certificate and certificate validation within the Perforce module. This way we can set an explicit dependency between the validation and the listener since they'd be within the same module. As with the other resources, we could conditionally create these depending on variables. 
2. We could also consider adding a `null_resource` or something similar that has a `wait` to help reduce the risk of this race condition. The challenge is that this validation is not a fixed amount of time. So we likely would have to lean on the high side to ensure it resolves the issue, which would cause unnecessarily longer runs of `terraform apply`.

modules/perforce/lb.tf
```diff
@@ -0,0 +1,339 @@
...
+    principals {
+      type = "AWS"
+      # Allow the ELB service account to create the logs
+      identifiers = [data.aws_elb_service_account.main.id]
+    }
+    resources = [
+      # If user provides existing S3 bucket to user for logging, grant access to that bucket, and the specific path they provide. Otherwise, grant access to the bucket that the module will conditionally create, and the optional user provided path or the default path defined
+      "${var.shared_lb_access_logs_bucket != null ? "${var.shared_lb_access_logs_bucket}/${var.shared_alb_access_logs_prefix}" : aws_s3_bucket.shared_lb_access_logs_bucket[0].arn}/${var.shared_alb_access_logs_prefix != null ? var.shared_alb_access_logs_prefix : "${var.project_prefix}-perforce-shared-alb"}/*",
+
+      # If user provides existing S3 bucket to user for logging, grant access to that bucket, and the specific path they provide. Otherwise, grant access to the bucket that the module will conditionally create, and the optional user provided path or the default path defined
```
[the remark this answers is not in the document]
novekm: ah good catch, will add to final commit which will contain the last remaining changes required for PR approval

modules/perforce/main.tf
novekm: Have time to huddle about this? I was a bit confused on how P4 server uses this value. From what I gather, you mean this should be `"https://${var.p4_auth.fully_qualified_domain_name}"`? Will be easier for me to resolve with you on a call to understand what you mean more clearly

modules/perforce/main.tf
novekm: yep, for the conditional data source stuff mentioned above

modules/perforce/main.tf
henrykie: You are instantiating the P4 Server and telling it where to find the Authentication server.

`var.p4_server_config.fully_qualified_domain_name` should be the fully qualified domain name of the p4 server. This is what you pass on line 13.

The `auth_server_url` should be the URL where the auth service can be found... I don't know why this would work for you if you were setting it to itself. Perforce would deploy, but the extension would be misconfigured. Check `/var/log/p4_configure.log` again - I almost guarantee that the extension fails to setup.

henrykie (dismissed):

modules/perforce/main.tf
novekm: Ah, I understand now. I was able to successfully reach the web page for`auth.perforce.<my-domain>` even with the other url (`perforce.<my-domain>`) set in the module config, so I assumed it was working. But from what I gather from what you mentioned, that is because the P4 Auth **_web service_** itself it working. However, if I actually tried to do anything with the service (such as login to P4 server with a user through p4 auth), that's what wouldn't work?

When I log into the container for p4 auth and run `cat /var/log/p4_configure.log` I get this response:
```
# cat /var/log/p4_configure.log
cat: /var/log/p4_configure.log: No such file or directory
```
I think this is because of the entrypoint. I had to modify it to be `cat ../var/log/p4_configure.log` but it still failed and returned: 
```
# cat /var/log/p4_configure.log
cat: /var/log/p4_configure.log: No such file or directory
```

If I look inside `/var/log/` the only files/directories I see are:
```
alternatives.log  amazon  apt  btmp  dpkg.log  faillog  lastlog  wtmp
```

So it may not have done the config for some reason? Not sure. Either way, I'll change the url to be the correct one and re-apply. Maybe that will fix

novekm: Ah sorry, just realized that you meant to run that on the P4 EC2 instance. Since that AMI is built using packer, I'm assuming the `p4_configure.sh` script only would have ran once, so even if I changed to the new url, it wouldn't reflect in the instance unless I rebuilt the AMI? If so, this is the log generated during the initial apply (with the incorrect `auth_service_url`):
```sh
sh-5.2$ cat /var/log/p4_configure.log
Thu Apr 24 00:09:05 UTC 2025 - Starting the p4 configure script.
Thu Apr 24 00:09:05 UTC 2025 - EBS_LOGS: /dev/sdf
Thu Apr 24 00:09:05 UTC 2025 - EBS_METADATA: /dev/sdg
Thu Apr 24 00:09:05 UTC 2025 - EBS_DEPOTS: /dev/sdh
Thu Apr 24 00:09:05 UTC 2025 - CASE_SENSITIVE: 1
Thu Apr 24 00:09:05 UTC 2025 - UNICODE: false
Thu Apr 24 00:09:05 UTC 2025 - SELINUX: false
Thu Apr 24 00:09:05 UTC 2025 - PLAINTEXT: false
Thu Apr 24 00:09:08 UTC 2025 - Attempt 1: One or more required paths are not valid EBS volumes or FSx mount points.
Thu Apr 24 00:09:13 UTC 2025 - Performing operations for mounting and syncing directories.
Thu Apr 24 00:09:13 UTC 2025 - Creating XFS file system on /dev/sdf.
Thu Apr 24 00:09:14 UTC 2025 - Mounting /dev/sdf on /mnt/temp_hxlogs.
Thu Apr 24 00:09:14 UTC 2025 - Creating XFS file system on /dev/sdg.
Thu Apr 24 00:09:14 UTC 2025 - Mounting /dev/sdg on /mnt/temp_hxmetadata.
Thu Apr 24 00:09:14 UTC 2025 - Creating XFS file system on /dev/sdh.
Thu Apr 24 00:09:15 UTC 2025 - Mounting /dev/sdh on /mnt/temp_hxdepots.
Thu Apr 24 00:09:17 UTC 2025 - Mounting /dev/sdf on /hxlogs.
Thu Apr 24 00:09:17 UTC 2025 - Mounting /dev/sdg on /hxmetadata.
Thu Apr 24 00:09:17 UTC 2025 - Mounting /dev/sdh on /hxdepots.
Thu Apr 24 00:09:17 UTC 2025 - Operation completed successfully.
Thu Apr 24 00:09:17 UTC 2025 - /home/ec2-user/gpic_scripts/p4_configure.sh
Thu Apr 24 00:09:17 UTC 2025 - Starting the configuration part after mounting was done later will configure the commit or replica depending on configuration.
Thu Apr 24 00:09:17 UTC 2025 - Updated P4ADMINPASS in /hxdepots/sdp/Server/Unix/setup/mkdirs.cfg.
Thu Apr 24 00:09:17 UTC 2025 - Updated ADMINUSER in /hxdepots/sdp/Server/Unix/setup/mkdirs.cfg.
Thu Apr 24 00:09:17 UTC 2025 - Updated P4MASTERHOST to ip-10-0-1-182.ec2.internal in /hxdepots/sdp/Server/Unix/setup/mkdirs.cfg.
Thu Apr 24 00:09:17 UTC 2025 - Updated CASE_SENSITIVE in /hxdepots/sdp/Server/Unix/setup/mkdirs.cfg.
Thu Apr 24 00:09:17 UTC 2025 - Skipping SSL_PREFIX removal from /hxdepots/sdp/Server/Unix/setup/mkdirs.cfg. Server will be configured to use SSL.
Thu Apr 24 00:09:17 UTC 2025 - Mounting done ok - continue to the install
Thu Apr 24 00:09:23 UTC 2025 - Generating self signed certificate
Thu Apr 24 00:09:27 UTC 2025 - Skipping SELinux label update
Thu Apr 24 00:09:27 UTC 2025 - Waiting for p4d_1 to start... Attempt 1 of 10.
Thu Apr 24 00:09:27 UTC 2025 - Service p4d_1 started successfully.
Thu Apr 24 00:09:43 UTC 2025 - Setting up SiteTags for installation
Thu Apr 24 00:09:43 UTC 2025 - Added awsuseast1 as a site tag
Thu Apr 24 00:09:43 UTC 2025 - Created SiteTags file appended AWS Region of this instance
Thu Apr 24 00:09:43 UTC 2025 - Configuring Helix Authentication Extension against https://perforce.novekm.people.aws.dev
Thu Apr 24 00:09:43 UTC 2025 - Starting Helix Authentication Extension setup.
Summary of arguments passed:

Helix server P4PORT            [ssl:1666]
Helix super-user               [perforce]
Service base URL               [https://perforce.novekm.people.aws.dev]
Preferred auth protocol        [oidc]
Debug logging enabled          [yes]
List of non-SSO users          [perforce]
List of non-SSO groups         [(not specified)]
List of SSO users              [(not specified)]
List of SSO groups             [(not specified)]
Name identifier property       [email]
Perforce user property         [email]

For a list of other options, type Ctrl-C to exit, and run this script with
the --help option.


The script is ready to make the configuration changes.

The operations involved are as follows:

  * Set global Service-URL to https://perforce.novekm.people.aws.dev
  * Set global Auth-Protocol to 'oidc'
  * Set instance enable-logging to 'on'
  * Set instance non-sso-users to perforce
  * Set instance name-identifier to email
  * Set instance user-identifier to email

installing new extension...
configuring global settings...
configuring instance settings...

The configure script will now run some tests to ensure that the extension
is functioning as expected...

Service error: 0: [CURL-EASY][COULDNT_CONNECT] Could not connect to server (7)
Service error: 0: [CURL-EASY][COULDNT_CONNECT] Could not connect to server (7)
Command invoke: OK

If every test indicates "OK" then the extension is ready.
If not, correct the issues before restarting the server.

Restarting Helix Core server...

==============================================================================
Automated configuration complete!

What was done:
  * The extension configuration was updated.
  * The Helix Core server was restarted.

What should be done now:
  * If not already completed, please install and configure the Helix
    Authenication Service on a system addressed by the URL entered in the
    extension configuration. If the URL must be changed in the extension
    configuration, you may change the value using the p4 extension command
    like so, and changing the Service-URL setting:

    $ p4 extension --configure Auth::loginhook

==============================================================================

Thu Apr 24 00:09:44 UTC 2025 - Skipping Unicode configuration
Thu Apr 24 00:09:44 UTC 2025 - EC2 mount script finished.
```

Is this the error you mean?
```sh
Service error: 0: [CURL-EASY][COULDNT_CONNECT] Could not connect to server (7)
Service error: 0: [CURL-EASY][COULDNT_CONNECT] Could not connect to server (7)
```

novekm: Confirmed, error resolved after updating the url

modules/perforce/modules/p4-server/variables.tf
novekm: Gotcha, reverting to `x86_64` and `c6i.large`

.gitignore
```diff
@@ -12,6 +12,10 @@
 crash.log
 crash.*.log
 
+# Ignore all directories that start with "autogen" anywhere in the project.
+# This is to ignore files that are automatically created using local_file within the Terraform modules
+autogen*/
```
novekm, line 17: we can prob remove this, but not a blocker

modules/perforce/modules/p4-server/versions.tf
```diff
@@ -0,0 +1,30 @@
+terraform {
```
novekm, line 1: I think we should remove all of the provider configs in the submodules. From the [TF Docs](https://developer.hashicorp.com/terraform/language/modules/develop/providers):

>Terraform v0.11 introduced the mechanisms described in earlier sections to allow passing provider configurations between modules in a structured way, and thus we explicitly recommended against writing a child module with its own provider configuration blocks.

modules/perforce/versions.tf
```diff
@@ -0,0 +1,26 @@
+terraform {
```
novekm, line 1: I think this should be updated to include the NetApp provider, right? It should be defined here, and not in the p4_server submodule `versions.tf` file

novekm (commented):
Looks good, just 2 small things regarding the provider config

PR #587, opened 2025-04-24
fix(helix swarm): remove helix_swarm_desired_container_count variable from module

### LINKED ISSUE #573: Bug: Helix Swarm fails on some containers when desired_container_count is greater than 1
### Expected Behaviour

The helix-swarm module supports setting the `desired_container_count` variable but setting it to anything but 1 breaks things. Either the variable should not be there, or setting it to a value higher than 1 should be supported.

### Current Behaviour

When setting `desired_container_count` to a value greater than 1, Swarm communication fails `desired_container_count - 1 / desired_container_count` of the times with:

> ERROR: Swarm communication error (Missing or invalid token. View 'About Swarm' as a super user for a list of valid tokens.)

### Possible Solution

_No response_

### Steps to Reproduce

1. Set the `desired_container_count` value to something greater than 1
2. Deploy
3. Use Perforce and check if everything works

Note: this bug was reported to me by e-mail. I haven't reproduced it, nor do I have a concrete reproduction scenario.

latest

jorisdon:
**Issue number:** fixes #573 

### Changes

This PR removes the `helix_swarm_desired_container_count` variable from the Helix Swarm module, as Helix Swarm does not scale horizontally, and any value greater than 1 would not work.

### User experience

Before: users could configure Helix Swarm in such a way that it wouldn't work - by setting desired_container_count to a value higher than 1. After: users can't configure Helix Swarm incorrectly.

## Checklist

* [ ] Changes are documented

<details>
No - although users will need to remove the `helix_swarm_desired_container_count` argument from their `perforce_helix_swarm` module if they've set one.
</details>

## Acknowledgment

### DIFF
--- a/modules/perforce/helix-swarm/variables.tf
+++ b/modules/perforce/helix-swarm/variables.tf
@@ -92,12 +92,6 @@ variable "existing_redis_connection" {
   default     = null
 }
 
-variable "helix_swarm_desired_container_count" {
-  type        = number
-  description = "The desired number of containers running the Helix Swarm service."
-  default     = 1
-}
-
 # - Existing Cluster -
 variable "cluster_name" {
   type        = string
--- a/modules/perforce/helix-swarm/main.tf
+++ b/modules/perforce/helix-swarm/main.tf
@@ -176,7 +176,7 @@ resource "aws_ecs_service" "helix_swarm_service" {
   cluster                = var.cluster_name != null ? data.aws_ecs_cluster.helix_swarm_cluster[0].arn : aws_ecs_cluster.helix_swarm_cluster[0].arn
   task_definition        = aws_ecs_task_definition.helix_swarm_task_definition.arn
   launch_type            = "FARGATE"
-  desired_count          = var.helix_swarm_desired_container_count
+  desired_count          = 1 # Helix Swarm does not scale horizontally, so desired container count is fixed at 1
   force_new_deployment   = var.debug
   enable_execute_command = var.debug
 


modules/perforce/helix-swarm/variables.tf
```diff
@@ -92,12 +92,6 @@ variable "existing_redis_connection" {
   default     = null
 }
 
-variable "helix_swarm_desired_container_count" {
-  type        = number
-  description = "The desired number of containers running the Helix Swarm service."
-  default     = 1
-}
```
jorisdon, line 99: Instead of outright removing this variable, would there be a reason for allowing it and constraining it to be between 0 and 1 inclusive? That way users could still retain the module, but remove the actual container deployment.

I'd argue this makes no sense and users should remove the usage of the module instead to disable Helix Swarm, guaranteeing support in the future for a module with `helix_swarm_desired_container_count` set to 0 is difficult and time consuming, plus such a configuration would result in a nonworking state unless Helix Swarm is not actually used.

henrykie: I think we just remove the variable. @novekm can ensure that the consolidated Perforce module conditionally creates the module and does not depend on the container count.

henrykie (approved):
LGTM

henrykie:
There is a checkov failure related to an LB listener... @novekm is this the same thing you saw?

```
Check: CKV2_AWS_74: "Ensure AWS Load Balancers use strong ciphers"
	FAILED for resource: aws_lb_listener.perforce_web_services_alb
```

novekm:
> There is a checkov failure related to an LB listener... @novekm is this the same thing you saw?
> 
> ```
> Check: CKV2_AWS_74: "Ensure AWS Load Balancers use strong ciphers"
> 	FAILED for resource: aws_lb_listener.perforce_web_services_alb
> ```

Yep, I had to add suppression

modules/perforce/helix-swarm/variables.tf
novekm: Does auth service support scaling or should we remove it there too?

henrykie:
I am overriding the checkov issue related to the listener since it seems to be a new Checkov rule.

Issue #513, opened 2025-02-27
Docs: Wrong Default Region in Getting Started Page


ltang-aws:
### What were you searching in the docs?

https://aws-games.github.io/cloud-game-development-toolkit/latest/getting-started.html

### How can we improve?

Perforce Helix Core no longer defaults to us-west-2, but will follow user configured CLI region to deploy

### Acknowledgment

- [x] I understand the final update might be different from my proposed suggestion, or refused.

awsbilal:
The other build agents in Step 3. Create Build Agent Amazon Machine Images, also should be set to have them deploy in the user's CLI configured region by default.  Testing those changes now.

PR #588, opened 2025-04-25
Linux Packer scripts changed to use Default Region

awsbilal:
#513 

### Changes

Linux based Packer scripts changed to use the user's CLI default region by default. 
Adjusted Getting Started.md documentation to reflect this. 
Also adjusted documentation to include unreferenced Packer scripts for Amazon Linux 2023 and Ubuntu Jammy.

### User experience

Before: Users would have linux based Packer scripts running in us-west-2 region even if their default region is somewhere else, and may have difficulty changing this behavior with the region variable manually. Users also may not have been aware that there were additional Linux scripts they could utilize, like `amazon-linux-2023-x86_64.pkr.hcl` and `ubuntu-jammy-22.04-arm64-server.pkr.hcl`

After: Now the user will not have to worry about setting the Region variable to configure the Linux Packer scripts to run in their CLI default regions. Also now aware of the additional Linux Packer scripts which are available to them.

## Checklist

* [x] Changes are documented

<details>

</details>

## Acknowledgment


henrykie (approved):

cwwalb (approved):
Looks good

Issue #533, opened 2025-03-13
Maintenance: Make TeamCity Module more readable


cwwalb:
### Why is this needed?

The main.tf file is big and should be split up by resource to improve readability

### Solution

The main.tf file is big and should be split up by resource to improve readability

PR #589, opened 2025-04-30
Maintenance: Split teamcity/main.tf into separate components

awsbilal:
**Issue number: #533 **

## Summary

Split main.tf module for teamcity into separate components to enhance readability.

### Changes

Created new .tf files, split by service or area, and removed unneeded main.tf file.

### User experience

before: Users that wanted to customize the terraform template file would have to o so from main.tf.

after: Users can find and customize different components in a more accessible manner.

## Checklist

* [x] Changes are documented

<details>

</details>

## Acknowledgment


ayatanb:
We split `main.tf` by following the patterns we saw in the other modules. There were logs set up to be stored in s3 that we put in a `logs.tf.` Can someone look over our new structure and let us know if how we split up `main.tf` is more readable per issue #533 ?

novekm:
I think we need to align on a standard for all of the modules in the project to ensure consistency. The possible options I can think of are:
1. Typical structure of a single `main.tf`, `variables.tf`, `outputs.tf`, and `providers.tf` (optionally `locals.tf`)
2. A `main.tf` file that contains the primary function of the module, additional `*.tf` files for related components 
3. No `main.tf` file, instead using service-specific names such as `ecs.tf`, `lb.tf`, `s3.tf`, etc
4. No `main.tf` file, instead using general functional use names such as `compute.tf`, `storage.tf`, `logging.tf`, `networking.tf`, etc

A main challenge with option 3 (which is what this module seems to do) is readability and consistency of where things should live within the files. While file segmentation of resources has the goal of increased readability, depending on how its implemented it can actually have the opposite effect. For example, if a load balancer needs logging, should that go in an `s3.tf` file since that would be the storage location for the logs, or should it be in the `alb.tf` file since it is used by the load balancer? Or as mentioned above, if it's anything to do with logging, should it go in a `logging.tf` file? It can cause a lot of flipping around between files depending on where these resources live.

I think we should work backwards from what the core component(s) of the module are, so essentially option 2 mentioned above. This is the approach I took for #585. Perhaps we could adopt a tenent of something like "if the file is longer than 300 lines, it probably could be split out a bit, but I hesitate to be that heavy handed. 

Most Terraform users will expect at the minimum a module to have a `main.tf` file that aligns with this. For this module I'm assuming that would be the ECS Task Definition, ECS Service, ECS Cluster, and EFS. From there, the additional files such as `lb.tf`, `iam.tf` and `sg.tf` could exist. What does everyone else think?

henrykie:
@kylesomers I want to hear your thoughts on this too... Kevon's opinion is one I share.

I think a `main.tf` is worth having, but when we look at optional related resources I think separate files make sense. For example, a specific `docdb` file for Horde since a user could provide an existing mongo connection string. Same goes for elasticache related work - valkey vs. redis vs. pre-existing dynamic configuration should have its own file. For TeamCity this would place the aurora related work and EFS related work in separate files. However, IAM roles or SGs related to any of these services might benefit from being alongside them. DB IAM role and SG inside of the `docdb.tf` or `rds.tf` files. Or maybe we just call it `database.tf`?

ayatanb:
> I think we need to align on a standard for all of the modules in the project to ensure consistency. The possible options I can think of are:
> 
> 1. Typical structure of a single `main.tf`, `variables.tf`, `outputs.tf`, and `providers.tf` (optionally `locals.tf`)
> 2. A `main.tf` file that contains the primary function of the module, additional `*.tf` files for related components
> 3. No `main.tf` file, instead using service-specific names such as `ecs.tf`, `lb.tf`, `s3.tf`, etc
> 4. No `main.tf` file, instead using general functional use names such as `compute.tf`, `storage.tf`, `logging.tf`, `networking.tf`, etc
> 
> A main challenge with option 3 (which is what this module seems to do) is readability and consistency of where things should live within the files. While file segmentation of resources has the goal of increased readability, depending on how its implemented it can actually have the opposite effect. For example, if a load balancer needs logging, should that go in an `s3.tf` file since that would be the storage location for the logs, or should it be in the `alb.tf` file since it is used by the load balancer? Or as mentioned above, if it's anything to do with logging, should it go in a `logging.tf` file? It can cause a lot of flipping around between files depending on where these resources live.
> 
> I think we should work backwards from what the core component(s) of the module are, so essentially option 2 mentioned above. This is the approach I took for #585. Perhaps we could adopt a tenent of something like "if the file is longer than 300 lines, it probably could be split out a bit, but I hesitate to be that heavy handed.
> 
> Most Terraform users will expect at the minimum a module to have a `main.tf` file that aligns with this. For this module I'm assuming that would be the ECS Task Definition, ECS Service, ECS Cluster, and EFS. From there, the additional files such as `lb.tf`, `iam.tf` and `sg.tf` could exist. What does everyone else think?

The challenge you mentioned is what Bilal and I were worried about and kind of wanted some insight on. We put the CloudWatch setup and the logs (creation of S3 bucket) in the `logs.tf`, and we weren't sure whether that was set up correctly. 

Also, where I'm confused on comes from the scope of `main.tf`. The readability is definitely an issue but if we require a `main.tf`, it would need ECS, EFS, and the Aurora database (like necessities to deploy the TeamCity server), which happens to be a majority of the code. Does that still improve the readability?

On the other hand, if we only have ECS and the task definitions in `main.tf`, shouldn't it just be called `ecs.tf`?

henrykie:
Let's add the "closes" keyword before the issue number here as well. I'm going to ask @novekm to help formalize this recommendation. Do we want to break a `main.tf` into multiple files? Do we want to break down by service (I don't think so)? Do we want to separate locals out, and if so should it be `local.tf` or `locals.tf`?

awsbilal:
So, to revive the discussion back here, do you guys think deciding on one of the existing modules as the "golden structure" and then just slowly working to convert other modules over to match it's structure would be the way to go?

novekm:
I propose we standardize with the following for all modules:
- `locals.tf` instead of `local.tf`
- create a `main.tf` that contains the core function of the module. For example, the P4 Auth submodule makes an ECS Cluster (optionally) and also a related task definition and service. Those are in `main.tf` and the other related resources (load balancer/target groups, IAM roles/policies, etc) are in separate files such as `iam.tf`, `lb.tf`, `sg.tf`, etc. I think we should use the new Perforce Module as the reference for both standalone modules, and modules that use submodules. We should create GitHub issues related to refactoring these modules in the manner
- use the `aws_iam_policy_document` data source instead of `jsonencode()` anytime an iam policy needs to be created

henrykie:
@ayatanb @awsbilal please review Kevon's message above, and use that as a basis for this refactor. The `main.tf` for the TC module should create ECS, Aurora and EFS resources. `iam.tf` should contain policies and role definitions. `sg.tf` should manage security group rules. I'd probably argue that the load balancers could be included in `main.tf`, but for consistencies sake lets also create an `lb.tf` for them to mimic the submodule approach in Perforce.

novekm (commented):
Initial thoughts - this looks good, however I think we should prioritize #621 before this. That issue would address adding automated tests for the TeamCity module, which will make it easier to validate that these changes work, as well as other changes that will likely come in the future. Once that is completed and merged, we can easily test this PR by commenting `/run-teamcity-module-tf-tests` on this PR

awsbilal:
@novekm Hmm, I see your point. Do you think it'll be much of a lift to get #621 done? How was #604?

henrykie:
image(8)

PR #592, opened 2025-05-04
Update Perforce complete example with new cidr allowlist variable

kylesomers:
**Issue number:** N/A

### Changes

In the Perforce complete example, the documentation provides steps to add a CIDR for allowlisting access to the deployed Perforce Helix Core server and Perforce Web Services NLB. This approach adds additional manual work for users and is error prone as it relies on adding an ingress rule in the AWS console or modifying the terraform module with new resources. To simplify this, this PR proposes removing these options and replace the allowlisting with an optional variable that the user provides when the complete example is deployed.  The documentation for this has also been updated to reflect this update.

### User experience

If a user wants to allowlist a cidr range as part of the complete example, they specify this through a new variable:

```
variable "enable_private_access_perforce" {
  type = object({
    enabled = bool
    cidr    = string
  })
  default     = null
  description = "Enable private access to Perforce and specify allowlisted CIDR range."
  validation {
    condition     = var.enable_private_access_perforce.enabled == true && var.enable_private_access_perforce.cidr != null
    error_message = "If private access to Perforce is enabled, a CIDR range must be provided."
  }
}
```

Example usage:

```
enable_private_access_perforce = {
  enabled = true
  cidr = "x.x.x.x/32"
}
```

With this variable set, the security group ingress rules will be conditionally added:

```
# Grants access on HTTPS port for Helix Swarm and Helix Authentication
resource "aws_vpc_security_group_ingress_rule" "private_perforce_https_ingress" {
  count             = var.enable_private_access_perforce.enabled ? 1 : 0
  security_group_id = aws_security_group.perforce_network_load_balancer.id
  description       = "Enables private access to Perforce web services."
  from_port         = 443
  to_port           = 443
  ip_protocol       = "TCP"
  cidr_ipv4         = var.enable_private_access_perforce.cidr
}

# Grants access on Helix Core port
resource "aws_vpc_security_group_ingress_rule" "private_perforce_helix_core_ingress" {
  count             = var.enable_private_access_perforce.enabled ? 1 : 0
  security_group_id = aws_security_group.perforce_network_load_balancer.id
  description       = "Enables private access to Perforce Helix Core."
  from_port         = 1666
  to_port           = 1666
  ip_protocol       = "TCP"
  cidr_ipv4         = var.enable_private_access_perforce.cidr
}
```

## Checklist

* [ ] Changes are documented

<details>

</details>

## Acknowledgment

### DIFF
--- a/docs/modules/perforce/examples/complete.md
+++ b/docs/modules/perforce/examples/complete.md
@@ -41,33 +41,13 @@ reference architecture. The deployment steps below will get you up and running.
 4. Review the plan provided by the above command. When you are ready to deploy you can confirm by typing "yes" on the
    command line. Terraform will take a few minutes to provision everything. When it completes, you are ready to proceed
    with testing.
-5. By default, none of the deployed resources are available on the public internet. This is to prevent unintended
-   security violations. You can update the security group for the Perforce Network Load balancer through the console, or
-   add the following rules to the example configuration in [
-   `security.tf`](https://github.com/aws-games/cloud-game-development-toolkit/blob/main/modules/perforce/examples/complete/security.tf):
-
-    ```terraform
-   # Grants access on HTTPS port for Helix Swarm and Helix Authentication
-   resource "aws_vpc_security_group_ingress_rule" "private_perforce_https_ingress" {
-      security_group_id = aws_security_group.perforce_network_load_balancer.id
-      description = "Enables private access to Perforce web services."
-      from_port = 443
-      to_port = 443
-      ip_protocol = "TCP"
-      cidr_ipv4 = "<YOUR IP>/32"
-   }
-
-   # Grants access on Helix Core port
-   resource "aws_vpc_security_group_ingress_rule" "private_perforce_https_ingress" {
-      security_group_id = aws_security_group.perforce_network_load_balancer.id
-      description = "Enables private access to Perforce Helix Core."
-      from_port = 1666
-      to_port = 1666
-      ip_protocol = "TCP"
-      cidr_ipv4 = "<YOUR IP>/32"
-
-   }
-    ```
+5. By default, none of the deployed resources are available on the public internet. This is to prevent unintended access to your resources. The perforce example includes a variable `enable_private_access_perforce` that you can configure to allowlist a CIDR to access the deployed Helix Core, Helix Auth, and Helix Swarm servers:
+```
+enable_private_access_perforce = {
+  enabled = true
+  cidr = "x.x.x.x/32"
+}
+```
 6. You should now have access to your deployed resources. The URLs for Helix Swarm and Helix Authentication Service are
    provided as Terraform outputs and should be visible in your console after a successful deployment. The connection
    string for Helix Core is also provided as an output. Use the Helix Core CLI or the P4V application to connect to your
--- a/modules/perforce/examples/complete/README.md
+++ b/modules/perforce/examples/complete/README.md
@@ -90,6 +90,8 @@ To use this example, you must have an existing **public Amazon Route53 Hosted Zo
 | [aws_vpc_security_group_ingress_rule.perforce_helix_core_inbound_web_alb_https](https://registry.terraform.io/providers/hashicorp/aws/5.89.0/docs/resources/vpc_security_group_ingress_rule) | resource |
 | [aws_vpc_security_group_ingress_rule.perforce_nlb_inbound_helix_core](https://registry.terraform.io/providers/hashicorp/aws/5.89.0/docs/resources/vpc_security_group_ingress_rule) | resource |
 | [aws_vpc_security_group_ingress_rule.perforce_nlb_inbound_web_alb_https](https://registry.terraform.io/providers/hashicorp/aws/5.89.0/docs/resources/vpc_security_group_ingress_rule) | resource |
+| [aws_vpc_security_group_ingress_rule.private_perforce_helix_core_ingress](https://registry.terraform.io/providers/hashicorp/aws/5.89.0/docs/resources/vpc_security_group_ingress_rule) | resource |
+| [aws_vpc_security_group_ingress_rule.private_perforce_https_ingress](https://registry.terraform.io/providers/hashicorp/aws/5.89.0/docs/resources/vpc_security_group_ingress_rule) | resource |
 | [aws_availability_zones.available](https://registry.terraform.io/providers/hashicorp/aws/5.89.0/docs/data-sources/availability_zones) | data source |
 | [aws_route53_zone.root](https://registry.terraform.io/providers/hashicorp/aws/5.89.0/docs/data-sources/route53_zone) | data source |
 
@@ -97,6 +99,7 @@ To use this example, you must have an existing **public Amazon Route53 Hosted Zo
 
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
+| <a name="input_enable_private_access_perforce"></a> [enable\_private\_access\_perforce](#input\_enable\_private\_access\_perforce) | Enable private access to Perforce and specify allowlisted CIDR range. | <pre>object({<br/>    enabled = bool<br/>    cidr    = string<br/>  })</pre> | `null` | no |
 | <a name="input_root_domain_name"></a> [root\_domain\_name](#input\_root\_domain\_name) | The root domain name you would like to use for DNS. | `string` | n/a | yes |
 
 ## Outputs
--- a/modules/perforce/examples/complete/main.tf
+++ b/modules/perforce/examples/complete/main.tf
@@ -206,25 +206,3 @@ resource "aws_lb_listener" "perforce_web_services_alb" {
     target_group_arn = aws_lb_target_group.perforce_web_services.arn
   }
 }
-
-# Grants access on HTTPS port for Helix Swarm and Helix Authentication
-resource "aws_vpc_security_group_ingress_rule" "private_perforce_https_ingress" {
-  count             = var.enable_private_access_perforce.enabled ? 1 : 0
-  security_group_id = aws_security_group.perforce_network_load_balancer.id
-  description       = "Enables private access to Perforce web services."
-  from_port         = 443
-  to_port           = 443
-  ip_protocol       = "TCP"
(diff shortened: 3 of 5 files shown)


novekm:
I think I already added similar logic in #585, however in that I'm using the existing functionality to add additional security groups instead of a new variable. 

For user access to the P4 Resources I see two main options:
- **Public Connectivity**
  - Public connectivity is currently through Route53 Public Hosted Zone to route traffic for P4 Server (Helix Core) P4Auth (Helix Authentication Service) and P4 Code Review (Helix Swarm) depending on on the host (e.g. `perforce.<your-domain>`, `auth.perforce.<your-domain>`, or `review.perforce.<your-domain`. When routed to these resource, either a SG on the P4 Server is used, or a SG on the public NLB is used.
    - To allow public traffic to the P4 resources, this can either be done by modifying rules for the default P4 Server, and NLB Security Groups that the module creates (using the new `cidr_allowlist` variable you are proposing), OR by creating your own security group and adding that as an additional security group that is attached to these resources. This can be set with the `existing_security_groups` variable at the example level.
- **Private Connectivity**
    - Private connectivity is currently only used for internal routing between the ECS Services behind the private ALB that the public NLB forwards requests to (P4 Server, P4 Code Review) and allowing connectivity from these services to the P4 Server.
      - To allow for private connectivity, you can use a VPN (Client VPN, S2S VPN) or other means to allow private traffic into the VPC where the resources are deployed. You then will need to modify the SGs or the SG rules for the private ALB (to allow private connectivity from your private IP range. You could either add an additional SG rule to the private ALB and P4 Server default Security Groups using the new `cidr_allowlist` variable you are proposing, OR you could use the `existing_security_groups` variable to pass in your own security group and attach to the these resources. 
          - The inbound rule for either option would be either:
              -  The private IP CIDR range for AWS Client VPN. This is created when you create an AWS Client VPN - all user connections would dynamically be assigned a private IP within this range
          - The private IP CIDR range from your on prem network for S2S VPN
  - **NOTE:** If allowing for private connectivity, we instead would be using the Route53 Private Hosted zone that the module creates by default. This would use A records to properly route requests, as is done with the public hosted zone currently. 

The example in #585 does option 1, public connectivity using the `existing_vpc_security_groups` variable.

## 1. Fetch your public IP with data source, create security group using this
```hcl
resource "aws_security_group" "allow_my_ip" {
  name        = "allow_my_ip"
  description = "Allow inbound traffic from my IP"
  vpc_id      = aws_vpc.perforce_vpc.id

  tags = {
    Name = "allow_my_ip"
  }
}

data "http" "my_ip" {
  url = "https://api.ipify.org" # will dynamically update the SG rule if your public IP changes
}

resource "aws_vpc_security_group_ingress_rule" "allow_https" {
  security_group_id = aws_security_group.allow_my_ip.id
  description       = "Allow HTTPS traffic from Kevon."
  from_port         = 443
  to_port           = 443
  ip_protocol       = "tcp"
  cidr_ipv4         = "${chomp(data.http.my_ip.response_body)}/32"
}
resource "aws_vpc_security_group_ingress_rule" "allow_http" {
  security_group_id = aws_security_group.allow_my_ip.id
  description       = "Allow HTTP traffic from Kevon."
  from_port         = 80
  to_port           = 80
  ip_protocol       = "tcp"
  cidr_ipv4         = "${chomp(data.http.my_ip.response_body)}/32"
}

resource "aws_vpc_security_group_ingress_rule" "allow_icmp" {
  security_group_id = aws_security_group.allow_my_ip.id
  description       = "Allow ICMP traffic from Kevon."
  from_port         = -1
  to_port           = -1
  ip_protocol       = "icmp"
  cidr_ipv4         = "${chomp(data.http.my_ip.response_body)}/32"
}
resource "aws_vpc_security_group_ingress_rule" "allow_perforce" {
  security_group_id = aws_security_group.allow_my_ip.id
  description       = "Allow Perforce traffic from Kevon."
  from_port         = 1666
  to_port           = 1666
  ip_protocol       = "tcp"
  cidr_ipv4         = "${chomp(data.http.my_ip.response_body)}/32"
}
```

## 2. Reference this existing security group in the module declaration:
```hcl
module "terraform-aws-perforce" {
  source = "../../"
  # source = "aws-games/perforce/aws" # remote example
  # version = "v1.0.0"

  # ...additional configuration

  # Add security group for your IP to the NLB as an additional security group. 
  existing_security_groups           = [aws_security_group.allow_my_ip.id] 
}
```

In the module, this is handled using the `concat()` function in Terraform, which will add the default P4 Server SG to the P4 Server, as well as any other SGs that are provided (using `var.existing_security_group_ids`).

### `/m
```hcl
##########################################
# Perforce Network Load Balancer
##########################################
resource "aws_lb" "perforce" {
  count                            = var.create_shared_network_load_balancer != false ? 1 : 0
  name_prefix                      = var.shared_network_load_balancer_name != null ? 
  load_balancer_type               = "network"
  subnets                          = var.public_subnets
  
  # ...additional configuration
  
  security_groups                  = concat(var.existing_security_groups, [aws_security_group.perforce_network_load_balancer[0].id]) # <-- this
  }
```

So I think this logic will already be made available when #585 is merged. The question is, do we want users to allow access by modifying the rules of the default SG that the Perforce module creates, or do we want them to do this by just adding an existing security group?

I think we should lean into the former - it abstracts away this management from the module (so we don't have to dynamically loop through `n` number of security group rules a user may want to provide) and instead we just reference any additional security groups that they pass into the module. What do you think?

Related to this, I think we should add more examples to cover both of these use cases. Right now we have an example for public connectivity (through Route53 Public Hosted Zone) but this requires some additional configuration from users, especially if they are not using Route53 for DNS. I propose we add another example for private connectivity, which I think should simplify DNS resolution, since we don't need to be concerned with DNS validation anymore I think. All we would need to ensure is that the ALB has a HTTPS certificate. Honestly, we may not even need this at that point? Since the traffic would be private over a VPN, we could even use HTTP. Though we should likely keep HTTPS if it's not complicated to manage.

kylesomers:
I missed that you had addressed this in the other PR.

> I think I already added similar logic in https://github.com/aws-games/cloud-game-development-toolkit/pull/585, however in that I'm using the existing functionality to add additional security groups instead of a new variable.

What existing functionality are you referring to? The Perforce complete example does not include any inbound access to the NLB. The documentation states that if this is needed, the user should either add ingress rule terraform blocks (examples provides in the docs) or they must manually add the ingress rule in the AWS console.

This PR does not address private connectivity use cases and I'd like to keep the discussion focused specifically on the public access. All this PR is doing is taking the existing manual steps in the documentation and automating them through an an optional variable. The broader questions around private vs public connectivity for accessing perforce resources is best addressed in a different thread, and we should probably include automation for setting up a client VPN as well, but not this PR.

In either case, we need users explicitly provide a cidr range (I'll find this in your PR and leave feedback there). I don't think it is a good idea or a security best practice to infer the IP automatically using an IP lookup service via a data source. This could lead to non-deterministic behavior for example if a user is behind a NAT or using public wifi. Also, the terraform module may be deployed using a remote server (codebuild, jenkins, GHA) so the IP that gets allowlisted won't be the correct one. In most cases a user would want to be allowlisting a CIDR representing a NAT/WAN IP of their office/home. This should be explicit.

novekm:
> What existing functionality are you referring to? The Perforce complete example does not include any inbound access to the NLB. The documentation states that if this is needed, the user should either add ingress rule terraform blocks (examples provides in the docs) or they must manually add the ingress rule in the AWS console.

I mean the existing functionality of the `existing_security_groups` variable - I think this was in the module before (for the helix core one) so I think I just added the same functionality to the NLB. In the previous example the NLB was created outside of the module, in the example, but I moved this within the module. 


> In either case, we need users to explicitly provide a cidr range (I'll find this in your PR and leave feedback there). I don't think it is a good idea or a security best practice to infer the IP automatically using an IP lookup service via a data source. This could lead to non-deterministic behavior for example if a user is behind a NAT or using public wifi. Also, the terraform module may be deployed using a remote server (codebuild, jenkins, GHA) so the IP that gets allowlisted won't be the correct one. In most cases a user would want to be allowlisting a CIDR representing a NAT/WAN IP of their office/home. This should be explicit.

So in that example I was mainly using the data source because I was planning to use the example as the configuration source for the automated Terraform tests. That way we wouldn't have to explicitly pass in the value (since we wouldn't know the public IP of the remote server), and could ensure whichever remote server  the pipeline tests uses could access the P4 resources for simple tests to check for connectivity to the server and the ECS services to make sure they're running and accessible. 

I agree that users may want to do this a better way. What I mean is that the existing variable `existing_security_groups` will already let users attach multiple SGs. In #585 this is extended to allow it it to be used to attach additional SGs besides the default SG to the NLB (that the new module is now creating) as well as the P4 Server. I'd need to double check but I think I also added this logic to the private ALB. If not, I can add it in but may need to separate with `existing_public_security_groups` (for user provided SG with rules for public IPs, when configuring the resources for public access), and `existing_private_security_groups` (for user provided SG with rules for private IPs, when configuring the resources for private access) but I agree that's a different discussion. 

What I'm proposing is that we document that users should use create their own security groups outside of the module declaration and pass them in using the `existing_security_groups` variable that is applied to the relevant resources the module creates. That way they can add the IPs however they want - they're just adding it to their own security groups that they create outside of the module, and telling the module to also attach those security groups to the NLB and P4 Server. Same could be done for the ALB for private connectivity later on. 

Basically, I think we should let users pass in additional SGs that they manage instead of letting them add rules to SGs the module manages. That way we limit scope of what the module is responsible for. It also ensures that even if a user had to make a manual change (outside of Terraform) for some reason, it doesn't impact the state of the default SGs created within the module.

kylesomers:
> What I'm proposing is that we document that users should use create their own security groups outside of the module declaration and pass them in using the existing_security_groups variable that is applied to the relevant resources the module creates. 

@henrykie what is the intent of `existing_security_groups`? Were you the 

@novekm - All this PR is adding is ingress rules to the module's security group to enable user to access them publicly. Why create new security for this? That's extra steps for a user. What am I missing here? If the module gets destroyed, then so should the ingress rules.

henrykie:
Kevon's changes introduce a new SG in the `create-resources-complete` example that handles access to the deployed resources. I am closing this PR :)

PR #604, opened 2025-05-14
chore: Add Terraform tests for new Perforce module

### LINKED ISSUE #480: Terraform Test: Perforce default deployments
### Why is this needed?

Automated deployment that runs on all changes to the Helix modules validating configuration successfully deploys.

This issue covers the following:
- Existing VPC (prerequisite)
- Test provisions shared cluster for HAS and Swarm
- Test provisions shared ALB for HAS and Swarm
- Test provisions shared NLB for Helix Core and HAS / Swarm ALB
- Existing certificate + DNS (prerequisite)

### Which area does this relate to?

CI/GitHub Actions

novekm:
**Issue number:**
Closes #480 

## Summary
This PR adds tests for the new Perforce module by leveraging `terraform test` functionality. As mentioned in #480, currently this is scoped to only perform simple unit testing (with `terraform plan`) until Terraform is updated to allow for more robust error handling and retry logic for `terraform test`, especially where state is concerned.

### Changes

This PR adds a new workflow file and related Terraform test files that are specific to the new Perforce module, and will test all of the current example implementations of the Perforce module. The goal is that this will allow basic automated tests for the module as additional functionality is added to it. When new functionality is added, additional examples and tests should be created simultaneously. This will allow both contributors and project maintainers to test PRs in a more scalable manner, helping to ensure code changes for net new additions function properly, and do not break any existing functionality.

The tests currently are invoked by commenting on the PR `/run-perforce-module-tf-tests` which will run the tests using a targeted `working-directory`:
```yaml
    defaults:
      run:
        working-directory: ${{ github.workspace }}/modules/perforce/
```

Within the tests, SSM Parameters of necessary values (input variables the example configurations require) are referenced from the relevant internal AWS account that is being used for CI for the toolkit. For additional tests within this module or for any module, this should be standardized - additional parameters added to the CI account, and using a `setup` module within the test to fetch those values and use them within the defined tests. This prevents hardcoding any potential secret values.

For additional module tests, for now we can make additional test files within the modules, that aligns to the example being used for the test. I recommend this take the naming of `01_<example name>.tftest.hcl` such as:
- `/modules/teamcity/tests/01_simple.tftest.hcl`
- `/modules/unreal/horde/tests/01_complete.tftest.hcl` 
- etc

and related workflow files taking the naming of `tf-test-on-comment-modules-<module name>.yml` such as:
- `tf-test-on-comment-modules-teamcity.yml`
- `tf-test-on-comment-modules-horde.yaml`
- etc

This allows to scope tests to the the modules, and easily run all tests for a desired module. Some of this functionality can likely be reused to run **all tests for all modules** when cutting a new release of the toolkit.

### User experience

This change is transparent to end users, and only directly impacts contributors and project maintainers in the manner mentioned above.

## Checklist

* [ ] Changes are documented

<details>
No

</details>

## Acknowledgment

### DIFF
--- a/.github/workflows/tf-test-on-comment-modules-perforce.yml
+++ b/.github/workflows/tf-test-on-comment-modules-perforce.yml
@@ -13,12 +13,8 @@ jobs:
   terraform-testing:
     # TODO: scope these permissions down
     permissions: write-all
-    # permissions:
-    #   id-token: write # This is required for AWS OIDC connection
-    #   contents: read # This is required for actions/checkout
-    #   pull-requests: write # This is required for GitHub bot to comment on PR
     name: Terraform Testing
-    # Only run if it#s a PR and the comment contains '/run-perforce-module-tf-tests'
+    # Only run if it is a PR and the comment contains '/run-perforce-module-tf-tests'
     if: github.event.issue.pull_request && contains(github.event.comment.body, '/run-perforce-module-tf-tests')
     runs-on: ubuntu-latest
     environment: aws-ci
--- a/.github/workflows/tf-test-on-comment-modules-perforce.yml
+++ b/.github/workflows/tf-test-on-comment-modules-perforce.yml
@@ -1,0 +1,140 @@
+name: "Invoke Terraform test on Comment for Perforce module"
+
+# https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows
+on:
+  issue_comment:
+    types: [created]
+
+env:
+  TF_LOG: INFO
+  AWS_REGION: ${{vars.AWS_REGION}}
+
+jobs:
+  terraform-testing:
+    # TODO: scope these permissions down
+    permissions: write-all
+    # permissions:
+    #   id-token: write # This is required for AWS OIDC connection
+    #   contents: read # This is required for actions/checkout
+    #   pull-requests: write # This is required for GitHub bot to comment on PR
+    name: Terraform Testing
+    # Only run if it#s a PR and the comment contains '/run-perforce-module-tf-tests'
+    if: github.event.issue.pull_request && contains(github.event.comment.body, '/run-perforce-module-tf-tests')
+    runs-on: ubuntu-latest
+    environment: aws-ci
+
+    defaults:
+      run:
+        working-directory: ${{ github.workspace }}/modules/perforce/
+
+    steps:
+      # Comment on PR that the workflow has been scheduled
+      - name: Comment on PR that the workflow has been scheduled
+        uses: actions/github-script@v6
+        if: always()
+        with:
+          script: |
+            const name = '${{ github.workflow	}}';
+            const url = '${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}';
+            const body = `Terraform tests have been scheduled 🧪 \n${url}`;
+
+            await github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: body
+            })
+
+      # Get branch of PR
+      - name: Get branch of PR
+        uses: xt0rted/pull-request-comment-branch@v1
+        id: comment-branch
+
+      # Set latest commit status as pending
+      - name: Set latest commit status as pending
+        uses: myrotvorets/set-commit-status-action@master
+        with:
+          sha: ${{ steps.comment-branch.outputs.head_sha }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+          status: pending
+
+      # Comment on PR that the workflow is in progress
+      - name: Comment on PR that the workflow is in progress
+        uses: actions/github-script@v6
+        if: always()
+        with:
+          script: |
+            const name = '${{ github.workflow	}}';
+            const url = '${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}';
+            const body = `Terraform tests are in progress ⏳ \n${url}`;
+
+            await github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: body
+            })
+
+      # Checkout the PR branch
+      - name: Checkout PR branch ${{ steps.comment-branch.outputs.head_ref }}
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ steps.comment-branch.outputs.head_ref }}
+
+      # Use GitHub Action secret to assume existing AWS IAM Role using OIDC connection
+      - name: Configure AWS credentials from AWS account
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{secrets.AWS_CI_ROLE_ARN}}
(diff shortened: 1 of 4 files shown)


henrykie (changes requested):
Minor things! Otherwise looks good.

.github/workflows/tf-test-on-comment-modules-perforce.yml
```diff
@@ -0,0 +1,140 @@
...
+  AWS_REGION: ${{vars.AWS_REGION}}
+
+jobs:
+  terraform-testing:
+    # TODO: scope these permissions down
+    permissions: write-all
+    # permissions:
+    #   id-token: write # This is required for AWS OIDC connection
+    #   contents: read # This is required for actions/checkout
+    #   pull-requests: write # This is required for GitHub bot to comment on PR
```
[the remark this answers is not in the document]
novekm: Initially was trying to use scoped permissions so was doing them individually, but was running into issues. For now just added `write-all` so I'll probably just remove the comments and keep that for now

.github/workflows/tf-test-on-comment-modules-perforce.yml
```diff
@@ -0,0 +1,140 @@
...
+    name: Terraform Testing
+    # Only run if it#s a PR and the comment contains '/run-perforce-module-tf-tests'
```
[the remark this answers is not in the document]
novekm: I was thinking that initially, but settled on the PR comment as the trigger. The Perforce resources take a long time to create and destroy, so I didn't want the pipeline to be triggered every time there's a change. I wanted this to be more of a manual step, likely to be used initially when reviewing a PR. My thought is that the flow would be:
1. contributor submits PR
2. reviewer comments `/run-perforce-module-tf-tests` to kick off the terraform tests
3. while pipeline is running, they look into the files changes for any areas of concern
4. if pipeline tests succeed and no areas of concern in the proposed changes, PR can be merged. Optionally for now, we would want to still do a manual run of `terraform apply` outside the pipeline

Right now the test are only doing `terraform plan` so the create/delete time isn't an issue so we do the automatic run on any changes to `.tf` files in the mofule, but once the tests are expanded to handle `terraform apply` (once Terraform `v1.13` adds better error handling/retry logic) I think it may still lead to long pipeline runs

modules/perforce/tests/setup/ssm.tf
```diff
@@ -0,0 +1,24 @@
+# Fetch relevant values from SSM Parameter Store
+data "aws_ssm_parameter" "route53_public_hosted_zone_name" {
+  name = "/cloud-game-development-toolkit/modules/perforce/route53-public-hosted-zone-name"
+}
+data "aws_ssm_parameter" "fsxn_password" {
+  name = "/cloud-game-development-toolkit/modules/perforce/fsxn-password"
+}
+data "aws_ssm_parameter" "fsxn_aws_profile" {
+  name = "/cloud-game-development-toolkit/modules/perforce/fsxn-aws-profile"
+}
```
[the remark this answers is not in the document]
novekm: I was debating between these two options as well. What would it look like to pass in the GH environment secrets? Either way I like standardizing with the `setup` module for any helper resources we need to reference or create that the other tests would consume. I think it also helps make it a bit easier to troubleshoot, since you can't see the value of the GH secret after it was created. That may make it a bit harder to troubleshoot if it's something simple like the secret has the wrong value or something.

In the nature of the toolkit providing examples the customers can use for a variety of things, I think it also has additional value of providing an example of using a `setup` module for tests. Lately I've been opting to writing any value my TF code creates that I may need later to SSM Parameter Store so I can use it elsewhere (outside of Terraform). It was also easy for me to check the functionality of the test - I simply went into the parameter, could see the correct value (in this case the Route53 Public Hosted Zone name) and added a small typo that I knew would make the test fail

henrykie (approved):
LGTM.

.github/workflows/tf-test-on-comment-modules-perforce.yml
```diff
@@ -0,0 +1,136 @@
...
+  TF_LOG: INFO
+  AWS_REGION: ${{vars.AWS_REGION}}
+
+jobs:
+  terraform-testing:
+    # TODO: scope these permissions down
+    permissions: write-all
+    name: Terraform Testing
+    # Only run if it is a PR and the comment contains '/run-perforce-module-tf-tests'
+    if: github.event.issue.pull_request && contains(github.event.comment.body, '/run-perforce-module-tf-tests')
```
gabebatista, line 18: Since the conditional is being checked within a job does this mean that this workflow will be triggered every time a comment (any comment) is made on the issues? and I guess a more important questions is.... does it matter?

.github/workflows/tf-test-on-comment-modules-perforce.yml
```diff
@@ -0,0 +1,136 @@
...
+
+    defaults:
+      run:
+        working-directory: ${{ github.workspace }}/modules/perforce/
+
+    steps:
+      # Comment on PR that the workflow has been scheduled
+      - name: Comment on PR that the workflow has been scheduled
+        uses: actions/github-script@v6
+        if: always()
```
gabebatista, line 30: From what I understand steps within a GitHub Action job always runs sequentially. Since this is the first step on the first (and only) job in this file, wouldn't this step always run no matter what?

.github/workflows/tf-test-on-comment-modules-perforce.yml
```diff
@@ -0,0 +1,136 @@
...
+
+        # Initialize a new or existing Terraform working directory
+      - name: Terraform Init
+        id: init
+        run: terraform init
+
+      # Run the actual Terraform tests
+      - name: Terraform Test
+        id: test
+        run: terraform test
```
gabebatista, line 108: The way this is setup now means it will run all tests? I'm assuming we would be replicating this file for each module and filtering to their specific tests?

gabebatista (approved):
This looks good, I left you some questions but they are mostly for my understanding rather than actual issues.

.github/workflows/tf-test-on-comment-modules-perforce.yml
novekm: Nah, it'll only run when `/run-perforce-module-tf-tests` is commented on a PR. Ideally we want to eventually scope this so only a project maintainer that comments this will actually invoke the action

.github/workflows/tf-test-on-comment-modules-perforce.yml
novekm: Yeah I think you're right, we may be able to remove but I think it's fine to leave in for now

.github/workflows/tf-test-on-comment-modules-perforce.yml
novekm: So it'll run all tests, but only those defined within the context of the module. For example, this will first change directory into the `/modules/perforce` directory, then run all tests that are in the `/tests` directory. Right now there's just a single test, but ideally we'd want to test all of these within each module when a PR is submitted.

We'd have a different test file that targets the individual working directory (and related test files) for the other modules, invoked with different comments such as `/run-jenkins-tf-tests`, `/run-teamcity-tf-tests` etc

PR #626, opened 2025-06-12
feat: Unity Accelerator asset caching proxy

### LINKED ISSUE #461: Feature request: Unity Accelerator Terraform Module
### Use case

> The [Unity Accelerator](https://docs.unity3d.com/Manual/UnityAccelerator.html) is a caching proxy agent that keeps copies of a team’s imported assets to speed up teamwork. The goal of the accelerator is to help teams reduce iteration time. An Accelerator coordinates asset sharing when your team is working on the same local network so that you don’t need to reimport portions of your project. This is done via the [Asset Import Pipeline](https://docs.unity3d.com/Manual/AssetDatabase.html) v2 Unity Editor capabilities.

The Unity Accelerator can be deployed using Docker.](https://hub.docker.com/r/unity/accelerator)

### Solution/User Experience

``` terraform
module "unity_accelerator" {
    source = "/modules/unity/accelerator"
    vpc_id = "vpc-abcd1234"
    ...
    dashboard_username = "administrator"
    dashboard_password = "iL0v3Un!ty"
}
```

We should be able to provision multiple `unity_accelerator` modules and showcase latency based routing in a sample. Accelerators can also mirror shared cache data - decisions on what backing store to use for cache data need to be reviewed.

### Alternative solutions

_No response_

erikyb:
**Issue number:**
Closes #461 

## Summary
Creates the necessary infrastructure to set up and operate the asset caching proxy Unity Accelerator service.

### Changes

Introduces the Unity Accelerator module which includes:
- ECS cluster with Unity Accelerator tasks, based on the official Docker image
- EFS drive for persistent, shared storage
- External ALB to service HTTPS dashboard traffic, password-protected
- External  NLB to service object serialization via protobuf traffic
- Contains a sample for quick deployment

### User experience

Before:
- Team members repeatedly download the same large assets
- Long wait times when switching branches or updating projects
- Network congestion due to redundant downloads
- Significant time spent waiting for assets to import

After:
- Assets are cached locally on the first download, subsequent requests are served from the local cache
- Near-instant access to previously downloaded assets
- Smooth branch switching and project updates
- Reduced network traffic
- Minimal wait times for asset imports
- Seamless collaboration experience

## Checklist

* [X] Changes are documented

<details>
No.
</details>

## Acknowledgment


ltang-aws:
Easy to read and understand, looks great!

modules/unity/accelerator/examples/simple/dns.tf
```diff
@@ -0,0 +1,76 @@
+variable "root_domain_name" {
```
gabebatista: all variables should be moved to a dedicated `variables.tf` file for clarity

modules/unity/accelerator/examples/simple/dns.tf
```diff
@@ -0,0 +1,76 @@
...
+  }
+}
+
+# Create a certificate for Unity Accelerator dashboard
+resource "aws_acm_certificate" "unity_accelerator" {
+  domain_name       = "dashboard.unity.${data.aws_route53_zone.root.name}"
+  validation_method = "DNS"
+
+  tags = {
+    Environment = "test"
```
gabebatista: Tags shouldn't be hard coded. All your other tag blocks are correct, I think you just missed this one

modules/unity/accelerator/examples/simple/main.tf
```diff
@@ -0,0 +1,9 @@
+module "unity_accelerator" {
+  source                               = "../../"
+  vpc_id                               = aws_vpc.unity_accelerator_vpc.id
+  service_subnets                      = aws_subnet.private_subnets[*].id
+  lb_subnets                           = aws_subnet.public_subnets[*].id
+  alb_certificate_arn                  = aws_acm_certificate.unity_accelerator.arn
+  unity_accelerator_dashboard_username = "uauser"
+  unity_accelerator_dashboard_password = "admin"
```
gabebatista: Be careful about passing sensitive data as variables. These values are stored in plaintext on the Terraform state file. A better approach would be to have the user upload the secrets to AWS Secrets manager and pull it from there. Now, if the Unity Accelerator can be configured to require the user to generate a new password on first login, then this is fine.

modules/unity/accelerator/README.md
```diff
@@ -0,0 +1,248 @@
+# Unity Accelerator Module
+[Unity Accelerator](https://docs.unity3d.com/6000.0/Documentation/Manual/UnityAccelerator.html) is a caching service that speeds up project operations in [Unity](https://unity.com/) by storing and sharing asset cache data among team members, significantly reducing build times and asset imports by allowing team members to reuse previously processed assets instead of having to process the same assets independently.
+
+This Unity Accelerator deployment uses an [Elastic Container Service](https://aws.amazon.com/ecs/) cluster for task deployment, with Amazon [Elastic File System](https://aws.amazon.com/efs/) providing persistent storage for configurations and cache. Access is managed through two load balancers: an [Application Load Balancer](https://aws.amazon.com/elasticloadbalancing/application-load-balancer/) for secure, password-protected web dashboard access via HTTPS, and a [Network Load Balancer](https://aws.amazon.com/elasticloadbalancing/network-load-balancer/) for efficient cache-related protobuf traffic. Custom DNS configuration is implemented using two A records in the root hosted zone, ensuring seamless integration with existing infrastructure.
```
ghost: This README is great. One question: We implement custom DNS configuration via A records hosted in the root hosted zone, but doesn't this actually occur in `examples/simple`? Just want to make sure we're documenting things in the appropriate location.

modules/unity/accelerator/examples/simple/dns.tf
```diff
@@ -0,0 +1,76 @@
...
+# Create a record in the Hosted Zone for the Unity Accelerator dashboard
+resource "aws_route53_record" "unity_accelerator_dashboard" {
+  zone_id = data.aws_route53_zone.root.id
+  name    = "dashboard.unity.${data.aws_route53_zone.root.name}"
+  type    = "A"
+
+  alias {
+    name                   = module.unity_accelerator.external_alb_dns_name
+    zone_id                = module.unity_accelerator.external_alb_zone_id
+    evaluate_target_health = false
```
ghost: Best practice? Maybe we should shift to setting this as `true`?

modules/unity/accelerator/examples/simple/dns.tf
```diff
@@ -0,0 +1,76 @@
...
+}
+
+# Create a certificate for Unity Accelerator dashboard
+resource "aws_acm_certificate" "unity_accelerator" {
+  domain_name       = "dashboard.unity.${data.aws_route53_zone.root.name}"
+  validation_method = "DNS"
+
+  tags = {
+    Environment = "test"
+  }
```
ghost: We need to standardize on how we want to set tags for examples. @novekm any thoughts?

modules/unity/accelerator/examples/simple/main.tf
ghost: Do you want to pass these as plaintext? Or instruct the user to create a secret and pass an ARN that the module can pull down into the service on startup?

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+        {
+          containerPort = 80
+          hostPort      = 80
+          protocol      = "TCP"
+        },
+        {
+          containerPort = 443
+          hostPort      = 443
+          protocol      = "TCP"
+        },
```
ghost: Do we need to open 443 on the container if we default to ALB handling TLS termination? If ALB is listening on 443 and then forwarding to 80 is the container port 443 actually ever in use?

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+        {
+          containerPort = 10080
+          hostPort      = 10080
+          protocol      = "TCP"
+        },
+        {
+          containerPort = 10443
+          hostPort      = 10443
+          protocol      = "TCP"
+        }
```
ghost: Same as above. Does the NLB handle TLS termination? Or does it just pass through to the container?

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+      ]
+
+      environment = concat(local.base_env, local.password_env)
+
+      entryPoint = ["/bin/sh", "-c"]
+
+      command = [
+        <<-EOT
+        /usr/local/bin/unity-accelerator register adbv2 &&
+        /usr/local/bin/unity-accelerator dashboard password ${var.unity_accelerator_dashboard_username} --password '${var.unity_accelerator_dashboard_password}' &&
```
ghost: We could pass these via environment variables using the `secrets` block for the task definition. This might be more secure... Not sure if we care?

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+        /usr/local/bin/unity-accelerator dashboard list &&
+        /usr/local/bin/unity-accelerator run
+        EOT
+      ]
+
+      mountPoints = [
+        {
+          sourceVolume  = "agent-storage"
```
ghost: can you explain this volume name?

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+          containerPath = var.unity_accelerator_persist
```
ghost: Can we set this variable (`var.unity_accelerator_persist`) to anything? Or is the unity-accelerator application that you start above expecting a specific path?

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+  network_configuration {
+    subnets         = var.service_subnets
+    security_groups = [aws_security_group.unity_accelerator_service_sg.id]
+  }
+
+  # Dashboard
+  load_balancer {
+    target_group_arn = aws_lb_target_group.unity_accelerator_dashboard_target_group[0].arn
+    container_name   = var.container_name
+    container_port   = var.unity_accelerator_dashboard_port
```
ghost: These probably shouldn't be variables. I know we've set them as vars for other applications, but it seems to me like the unity-accelerator application is always going to run on port 80 and port 10080.

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+  }
+
+  # Cache
+  load_balancer {
+    target_group_arn = aws_lb_target_group.unity_accelerator_cache_target_group[0].arn
+    container_name   = var.container_name
+    container_port   = var.unity_accelerator_cache_port
```
ghost: same as above.

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+# ECS ingres rules to allow health checks from NLB subnets
+resource "aws_vpc_security_group_ingress_rule" "unity_accelerator_service_ingress_from_nlb_80" {
+  for_each          = var.create_external_nlb ? data.aws_subnet.nlb_subnets : {}
+  security_group_id = aws_security_group.unity_accelerator_service_sg.id
+  description       = "Allows inbound HTTP health checks from NLB"
+  from_port         = var.unity_accelerator_dashboard_port
+  to_port           = var.unity_accelerator_dashboard_port
+  ip_protocol       = "TCP"
+  cidr_ipv4         = each.value.cidr_block
+}
```
ghost: This is the second time I've seen this. Can we not use a referenced SG for the ingress rules from the NLB? @cwwalb I know you encountered this as well.

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+
+# Egress rule for NFS traffic from service to EFS
+resource "aws_vpc_security_group_egress_rule" "ecs_to_efs" {
+  security_group_id            = aws_security_group.unity_accelerator_service_sg.id
+  referenced_security_group_id = aws_security_group.unity_accelerator_efs_sg[0].id
+  description                  = "Allows outbound NFS traffic from Unity Accelerator service to EFS"
+  from_port                    = 2049
+  to_port                      = 2049
+  ip_protocol                  = "TCP"
+}
```
ghost: I think this is redundant since you also have a rule below that allows all outbound from the service

modules/unity/accelerator/examples/simple/vpc.tf
```diff
@@ -0,0 +1,123 @@
...
+
+  tags = merge(local.tags,
+    {
+      Name = "unity-accelerator-private-rt"
+    }
+  )
+}
+
+# route to the internet through NAT gateway
+resource "aws_route" "private_rt_nat_gateway" {
```
gabebatista, line 103: Does this solution need access to the open internet? or just to the developer's machines?

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+      enabled = var.enable_unity_accelerator_lb_access_logs
+      bucket  = var.unity_accelerator_lb_access_logs_bucket != null ? var.unity_accelerator_lb_access_logs_bucket : aws_s3_bucket.unity_accelerator_lb_access_logs_bucket[0].id
+      prefix  = var.unity_accelerator_alb_access_logs_prefix != null ? var.unity_accelerator_alb_access_logs_prefix : "${local.name_prefix}-alb"
+    }
+  }
+
+  enable_deletion_protection = var.enable_unity_accelerator_lb_deletion_protection
+  drop_invalid_header_fields = true
+  tags                       = local.tags
+}
```
ghost: I would prefer we have a variable: `var.create_application_load_balancer` and a variable: `var.internal_alb` to control these two things separately. If a user wants to leverage a shared ALB they can create that external to the module and use the target group output to connect the accelerator application to that load balancer. If a user wants to create an ALB as part of the module they use `var.create_application_load_balancer` (which defaults to true). Whether or not that load balancer is internal or external is a separate question.

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+
+resource "aws_lb_target_group" "unity_accelerator_dashboard_target_group" {
+  #checkov:skip=CKV_AWS_378: Using ALB for TLS termination
+
+  count       = var.create_external_alb ? 1 : 0
```
ghost: This should always be created. If the user is not creating an ALB _within_ the module they will still need the target group to connect from an externally created ALB.

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+resource "aws_lb" "unity_accelerator_external_nlb" {
+  #checkov:skip=CKV_AWS_150:Deletion protection disabled by default
+  #checkov:skip=CKV2_AWS_28:NLB access is managed with SG allow listing
+  #checkov:skip=CKV_AWS_152:Module is single-region
+
+  count              = var.create_external_nlb ? 1 : 0
+  name               = "${local.name_prefix}-nlb"
+  load_balancer_type = "network"
+  internal           = false
+  subnets            = var.lb_subnets
```
ghost: Same thing as above:
`var.create_network_load_balancer` and `var.internal_nlb` to control network configuration and conditional creation of the load balancers. Let me know if you have questions on this - happy to expand.

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+      prefix  = var.unity_accelerator_nlb_access_logs_prefix != null ? var.unity_accelerator_nlb_access_logs_prefix : "${local.name_prefix}-nlb"
+    }
+  }
+
+  enable_deletion_protection = var.enable_unity_accelerator_lb_deletion_protection
+  tags                       = local.tags
+}
+
+resource "aws_lb_target_group" "unity_accelerator_cache_target_group" {
+  count       = var.create_external_nlb ? 1 : 0
```
ghost: Should always be created.

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+# VPC Endpoints for ECS Execute Command
+resource "aws_vpc_endpoint" "ssm_vpce" {
+  count               = var.debug == true ? 1 : 0
+  vpc_id              = var.vpc_id
+  service_name        = "com.amazonaws.${data.aws_region.current.name}.ssm"
+  vpc_endpoint_type   = "Interface"
+  subnet_ids          = var.service_subnets
+  security_group_ids  = [aws_security_group.vpc_endpoint_sg[0].id]
+  private_dns_enabled = true
+}
```
ghost, line 787: This is a really good addition. We should standardize this. Can you create an issue in the project to do this across other modules?

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+  description = "Security group for VPC endpoints"
+  vpc_id      = var.vpc_id
+
+  ingress {
+    from_port       = 443
+    to_port         = 443
+    protocol        = "TCP"
+    security_groups = [aws_security_group.unity_accelerator_service_sg.id]
+    description     = "Allows HTTPS traffic from Unity Accelerator service to VPC endpoints"
+  }
```
ghost: I prefer controlling rules as a separate resource as you did previously

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+# Allow the ECS tasks to access the VPC endpoints
+resource "aws_vpc_security_group_egress_rule" "unity_accelerator_service_to_vpce" {
+  count                        = var.debug == true ? 1 : 0
+  security_group_id            = aws_security_group.unity_accelerator_service_sg.id
+  referenced_security_group_id = aws_security_group.vpc_endpoint_sg[0].id
+  from_port                    = 443
+  to_port                      = 443
+  ip_protocol                  = "TCP"
+  description                  = "Allow outbound HTTPS to VPC endpoints"
+}
```
ghost: I believe this is redundant since the service as an "all outbound" rule attached as well.

modules/unity/accelerator/outputs.tf
```diff
@@ -0,0 +1,19 @@
...
+
+output "external_nlb_dns_name" {
+  value       = var.create_external_nlb ? aws_lb.unity_accelerator_external_nlb[0].dns_name : null
+  description = "DNS endpoint of Network Load Balancer (NLB)"
+}
+
+output "external_nlb_zone_id" {
+  value       = var.create_external_nlb ? aws_lb.unity_accelerator_external_nlb[0].zone_id : null
+  description = "Zone ID for internet-facing Network Load Balancer (NLB)"
+}
```
ghost: With the addition of an `internal` variable to control LB internal or external nature we should update these names

modules/unity/accelerator/variables.tf
```diff
@@ -0,0 +1,252 @@
...
+# Unity Accelerator Configuration
+####################################################
+
+variable "unity_accelerator_docker_image" {
+  type        = string
+  description = "Docker image to use for Unity Accelerator."
+  default     = "unitytechnologies/accelerator:latest"
+}
+
+variable "unity_accelerator_persist" {
```
ghost: maybe rename "unity_accelerator_persisted_directory" or something along those lines?

modules/unity/accelerator/variables.tf
```diff
@@ -0,0 +1,252 @@
...
+variable "unity_accelerator_dashboard_username" {
+  description = "Username for the Unity Accelerator dashboard."
+  type        = string
+}
+
+variable "unity_accelerator_dashboard_password" {
+  description = "Password for the Unity Accelerator dashboard."
+  type        = string
+  sensitive   = true
+}
```
ghost: Again, maybe we support an ARN for a secret to these instead of plaintext? Can be a discussion. @gabebatista @kylesomers what do you think?

modules/unity/accelerator/variables.tf
```diff
@@ -0,0 +1,252 @@
...
+  type        = string
+  description = "The name of the Unity Accelerator container."
+  default     = "unity-accelerator"
+}
+
+variable "unity_accelerator_dashboard_port" {
+  type        = number
+  default     = 80
+  description = "The port to reach the Unity Accelerator dashboard."
+}
```
ghost: This might not _actually_ be a variable right? Or is this where we map the host port? I think adding this level of control might be unnecessary (even though we do it loads of other places) and we could just use a local?

modules/unity/accelerator/variables.tf
```diff
@@ -0,0 +1,252 @@
...
+  default     = 10080
+  description = "The port to reach the Unity Accelerator cache, used by the Unity client."
+}
+
+variable "desired_container_count" {
+  type        = number
+  description = "The desired number of containers running Unity Accelerator."
+  default     = 1
+  nullable    = false
+}
```
ghost: Is the unity accelerator application horizontally scalable?

modules/unity/accelerator/variables.tf
```diff
@@ -0,0 +1,252 @@
...
+variable "enable_unity_accelerator_lb_deletion_protection" {
+  type        = bool
+  description = "Enables deletion protection for the Unity Accelerator Application Load Balancer and Network Load Balancer. Defaults to true."
+  default     = true
+}
+
+######
+# ALB
+######
+variable "create_external_alb" {
```
ghost: maybe not "external" - instead `create_application_load_balancer` and `create_network_load_balancer`

henrykie:
Also, make sure to rebase main onto this branch to get it up to date!

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+  )
+}
+
+# ECS Service
+resource "aws_ecs_service" "unity_accelerator" {
+  name                   = local.name_prefix
+  cluster                = var.cluster_name != null ? data.aws_ecs_cluster.unity_accelerator_cluster[0].arn : aws_ecs_cluster.unity_accelerator_cluster[0].arn
+  task_definition        = aws_ecs_task_definition.unity_accelerator_task_definition.arn
+  launch_type            = "FARGATE"
+  desired_count          = var.desired_container_count
```
gabebatista: Does the accelerator support horizontal scaling or only failover?

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,817 @@
...
+  count       = var.efs_id == null ? 1 : 0
+  name        = "${local.name_prefix}-efs-sg"
+  description = "Unity Accelerator EFS mount target security group"
+  vpc_id      = var.vpc_id
+  tags        = local.tags
+}
+
+# Ingress rule for NFS traffic from service to EFS
+resource "aws_vpc_security_group_ingress_rule" "service_efs" {
+  count                        = var.efs_id == null ? 1 : 0
```
gabebatista, line 267: if `efs_id` is not present, we create an ingress rule?

modules/unity/accelerator/main.tf
gabebatista, line 1: This file is way too long, you may want to break off some of the services, like maybe a dedicated file for the load balancer? or EFS?

modules/unity/accelerator/variables.tf
```diff
@@ -0,0 +1,252 @@
...
+variable "unity_accelerator_log_stdout" {
+  type        = string
+  description = "When true, outputs logs to stdout only. When false, writes logs to the persist directory."
+  default     = "true"
+}
+
+variable "unity_accelerator_debug_mode" {
+  type        = string
+  description = "Enables debug output for the Unity Accelerator service."
+  default     = "true"
```
gabebatista: I would default this to false. Might be worth calling out that this is an option in the docs

modules/unity/accelerator/variables.tf
```diff
@@ -0,0 +1,252 @@
...
+  default     = true
+}
+
+variable "unity_accelerator_lb_access_logs_bucket" {
+  type        = string
+  description = "ID of the S3 bucket for Unity Accelerator Application Load Balancer and Network Load Balancer access log storage. If access logging is enabled and this is null the module creates a bucket."
+  default     = null
+}
+
+variable "enable_unity_accelerator_lb_deletion_protection" {
```
gabebatista, line 182: what happens if both this and debug are true? Might be worth creating a validation here to keep things clear for the user.

modules/unity/accelerator/variables.tf
gabebatista: Is the idea here that if someone wants to provide their own alb, they would make this false then manually connect their own? In other modules we allow users to pass in the arn of a pre-existing alb (assuming they want to use their own) and use that as a conditional to create our own albs or not.

gabebatista (changes requested):
Over all this looks good. I asked for some minor refactoring to make the code a little more "digestible", especially so our newer members can read through it too. There were also some minor comments around security. Let me know if you need help addressing any of it.

modules/unity/accelerator/examples/simple/dns.tf
gabebatista: I like the way we are doing it in the other modules. Use a variable to capture user tags, use locals for static module level tags and inline for resource specific then merge them all together.

modules/unity/accelerator/variables.tf
gabebatista: That's what I suggested on my review also. I do wish we had a good way to capture all these secrets without pushing the user to interact with secrets manager. Maybe a little CLI wizard?

modules/unity/accelerator/examples/simple/dns.tf
erikyb: This is taken directly from the TeamCity module, and is in the example flow. For the module, tags are captured via a var and merged.

modules/unity/accelerator/README.md
erikyb: You're right, it does occur in the example flow. Will remove.

modules/unity/accelerator/examples/simple/dns.tf
erikyb: Will change. Note that this behavior is taken directly from the TeamCity module, in case the change also needs to happen there.

modules/unity/accelerator/examples/simple/main.tf
erikyb: After talking to Gabe, will add a prerequisite for the user to set up a password in Secrets Manager, and pass the ARN instead.

modules/unity/accelerator/main.tf
erikyb: Makes sense to me, will remove them and hardcode it to the default ports instead.

modules/unity/accelerator/main.tf
erikyb: It can be set to anything, default is `/agent`. I will remove the variable as I don't think there is any compelling reason to make it customizable.

modules/unity/accelerator/main.tf
erikyb: Sure thing, will do after PR is approved and merged to main, to link to example in the issue.

modules/unity/accelerator/examples/simple/vpc.tf
erikyb: Needs to be able to pull the latest unity accelerator docker image.

modules/unity/accelerator/main.tf
erikyb: Good call, Unity recommends 1 unity accelerator per project so will be removing the desired container count variable.

modules/unity/accelerator/main.tf
erikyb: You can specify an existing EFS to use as the persistent storage layer for the solution. If `var.efs_id` was not specified, it means we are creating our own, including ingress/egress rules.

modules/unity/accelerator/main.tf
erikyb: I was told the new way was to have everything in one `main.tf` file. I can break it up again (originally started this way) but I'd like to make sure I won't get told by someone else to put everything back on a `main.tf` file before doing so 🙂

modules/unity/accelerator/variables.tf
erikyb: Correct. Followed the approach used by the TeamCity module.

modules/unity/accelerator/main.tf
gabebatista: Fair enough, who told you to put it all in main? Henry?

modules/unity/accelerator/main.tf
gabebatista: Got it, The variable name for me a little confused

modules/unity/accelerator/main.tf
erikyb: He mentioned the project was moving away from service-specific tf files, but maybe I misunderstood what he meant. I have no preference, just want to make sure I understand what the expectation is. Let me know how you want it and I'll get it done!

modules/unity/accelerator/examples/simple/vpc.tf
gabebatista: That makes sense, but the thing that concerns me is that after the initial usage of pulling the container image the NAT Gateway will sit around racking up costs even though its no longer needed. I know that this is how things are done in other modules (I checked Horde as an example) so don't worry about changing it now. Just want to call this out so it can be addressed in a future version.

modules/unity/accelerator/main.tf
gabebatista: I'm leaving this one as is and we can break down the `main.tf` file in a future version if necessary.

modules/unity/accelerator/examples/simple/vpc.tf
erikyb: Would having a `Notes` or `Additional Information` section and calling this out (and perhaps a couple other things) suffice? If I remove it at the end of the script when it's no longer needed, but for some reason they need to make changes to memory or something else and force a new deployment, it would fail.

modules/unity/accelerator/variables.tf
gabebatista: Looking through some of the other modules, it seems like this issue holds true for the rest of them. I will create a new issue to address this in a future version.

modules/unity/accelerator/variables.tf
```diff
@@ -0,0 +1,243 @@
...
+######
+# ALB
+######
+variable "create_alb" {
+  type        = bool
+  description = "Set this flag to true to create an Application Load Balancer for the Unity Accelerator dashboard."
+  default     = true
+}
+
+variable "internal_alb" {
```
gabebatista: The name of this variable makes it seem like if set to false we would not create the alb at all rather than making it external. Change the name to make its functionality more clear.

gabebatista (dismissed):
The module looks great, thank you for making the requested changes. I left one comment regarding a variable name, if you could just change that before we merge the branch we should be good to go (I didnt want to make you re-request a review for a variable name).

modules/unity/accelerator/examples/simple/main.tf
```diff
@@ -0,0 +1,9 @@
+module "unity_accelerator" {
+  source                                   = "../../"
+  vpc_id                                   = aws_vpc.unity_accelerator_vpc.id
+  service_subnets                          = aws_subnet.private_subnets[*].id
+  lb_subnets                               = aws_subnet.public_subnets[*].id
+  alb_certificate_arn                      = aws_acm_certificate.unity_accelerator.arn
+  unity_accelerator_dashboard_username     = "uauser"
+  unity_accelerator_dashboard_password_arn = "arn:aws:secretsmanager:us-east-1:112233445566:secret:77889900"
```
ghost: Neat trick: We can actually automate the creation of this password service side, and provide the ARN as an output. [Check out how the Perforce module does this.](https://github.com/aws-games/cloud-game-development-toolkit/blob/bc5119257e7c6130571ae147049553d08cda9a4b/modules/perforce/modules/p4-server/main.tf#L15-L20) The user can still provide an existing ARN for an existing secret, but if they don't the module creates a random password (which does not get persisted in TF state), and we can then create an `output` for that ARN so the user can go find the password or use the CLI to fetch its value.

ghost: Otherwise I'd recommend mandating this is passed as a variable. Otherwise a `terraform apply` would still run, and then only fail when that placeholder secret ARN cannot be found.

henrykie (changes requested):
Minor changes requested on the password secret management. Looks good to me!

henrykie (approved):
LGTM! 🚀

modules/unity/accelerator/main.tf
```diff
@@ -0,0 +1,863 @@
...
+
+# ECS ingress rule on port 10080 (cache)
+resource "aws_vpc_security_group_ingress_rule" "unity_accelerator_service_ingress_from_alb_10080" {
+  count             = var.create_alb ? 1 : 0
+  security_group_id = aws_security_group.unity_accelerator_service_sg.id
+  description       = "Allows HTTP traffic on port 10080 (cache)"
+  from_port         = 10080
+  to_port           = 10080
+  ip_protocol       = "TCP"
+  cidr_ipv4         = "0.0.0.0/0"
```
ayatanb: is it okay to have an ingress rule for 0.0.0.0?

ayatanb (dismissed):
otherwise looks solid to me! 👌

modules/unity/accelerator/variables.tf
ghost: A CLI wizard is just as much work as providing documentation that informs you on how to use the AWS CLI. Plus, its another thing that could break.

modules/unity/accelerator/main.tf
ghost: Is traffic to the cache service routed through NLB? If so, why aren't we attaching SG to NLB and using `referenced_security_group_id` to allow communications?

henrykie (approved):
LGTM! 🚀 

@gabebatista @danielw-aws perhaps a review?

gabebatista (approved):
LGTM! Great work Erik!

PR #634, opened 2025-06-19
Feature: Adding ECS Agents on TeamCity Server Startup

### LINKED ISSUE #635: Feature request: Create TeamCity agent on TeamCity server deployment
### Use case

This feature will add the capability to have TeamCity agents available in the TeamCity server when it is deployed.

### Solution/User Experience

TBD

### Alternative solutions

_No response_

ayatanb:
closes #635

## Summary

This PR adds the functionality of having TeamCity agents available in the TeamCity server on startup.

### Changes

- Adds `agents.tf` to the TeamCity module.
- `main.tf` will also have configurations for ECS Service Connect, which is being used for the communication between the TeamCity server and the agent(s).
- cleaned up the IAM policy document and attachments in `main.tf`.

### User experience

Users will optionally be able to deploy a TeamCity agent to the TeamCity server on startup, and can configure the following in the `build_farm_config` variable:

- the agent image
- desired count
- cpu and memory allocation for each task

The user can also create different types of agents within the `buiild_farm_config` in case different images need to be used.

## Checklist

* [ X ] I have performed a self-review of this change
* [ X ] Changes have been tested
* [ ] Changes are documented

<details>

no

</details>

## Acknowledgment

### DIFF
--- a/modules/teamcity/agents.tf
+++ b/modules/teamcity/agents.tf
@@ -1,0 +1,184 @@
+#################################
+###### ECS TASK AND SERVICE #####
+#################################
+
+resource "aws_ecs_task_definition" "teamcity_agent" {
+  # create a task for each build agent config in the build_farm_config variable
+  #checkov:skip=CKV_AWS_336: EFS is not necessary for TeamCity Agents
+
+  for_each = var.build_farm_config
+
+  family                   = each.key
+  network_mode             = "awsvpc"
+  requires_compatibilities = ["FARGATE"]
+  cpu                      = each.value.cpu
+  memory                   = each.value.memory
+  execution_role_arn       = aws_iam_role.teamcity_agent_task_execution_role.arn
+  task_role_arn            = aws_iam_role.teamcity_agent_default_role.arn
+  container_definitions = jsonencode([
+    {
+      name      = "teamcity-agent"
+      image     = each.value.image
+      cpu       = each.value.cpu
+      memory    = each.value.memory
+      essential = true
+      environment = [
+        {
+          name  = "SERVER_URL"
+          value = "http://teamcity-server.teamcity-namespace:8111"
+        },
+        {
+          name  = "AGENT_NAME"
+          value = "${each.key}-agent"
+        },
+
+      ]
+      logConfiguration = {
+        logDriver = "awslogs"
+        options = {
+          "awslogs-group"         = aws_cloudwatch_log_group.teamcity_agent.name
+          "awslogs-region"        = data.aws_region.current.name
+          "awslogs-stream-prefix" = "[AGENT - ${each.key}]"
+        }
+      }
+    }
+  ])
+}
+
+resource "aws_ecs_service" "teamcity_agent" {
+  for_each = var.build_farm_config
+
+  name = each.key
+  cluster = (var.cluster_name != null ? data.aws_ecs_cluster.teamcity_cluster[0].arn :
+  aws_ecs_cluster.teamcity_cluster[0].arn)
+  task_definition = aws_ecs_task_definition.teamcity_agent[each.key].arn
+  desired_count   = each.value.desired_count
+  launch_type     = "FARGATE"
+
+  network_configuration {
+    subnets         = var.service_subnets
+    security_groups = [aws_security_group.teamcity_agent_sg.id]
+  }
+
+  enable_execute_command = var.debug
+
+  service_connect_configuration {
+    enabled   = true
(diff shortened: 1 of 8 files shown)


modules/teamcity/main.tf
```diff
@@ -90,14 +102,19 @@ resource "aws_ecs_task_definition" "teamcity_task_definition" {
...
+  name = local.name_prefix
+  cluster = (var.cluster_name != null ? data.aws_ecs_cluster.teamcity_cluster[0].arn :
+  aws_ecs_cluster.teamcity_cluster[0].arn)
   task_definition        = aws_ecs_task_definition.teamcity_task_definition.arn
   launch_type            = "FARGATE"
   desired_count          = var.desired_container_count
   force_new_deployment   = var.debug
   enable_execute_command = var.debug
 
+  //trying to make it so that the databases dont break! temp solution
```
gabebatista: Can you elaborate a bit on what is happening and how this helps to address it?

modules/teamcity/variables.tf
cwwalb, line 1: the PR says that there is a ```build_farm_config``` variable. Was it changed to this one?

ayatanb: oh great catch, thanks Christina! I changed the name on local I might have forgotten to push that name change

joshua7459 (commented):
I am attempting to deploy the teamcity module, however after doing TF apply, i am asked to provide values for subnets and VPC IDs. After providing the values, the deployment errors and cancels itself
Screenshot 2025-07-10 at 8 42 46 PM

modules/teamcity/agents.tf
```diff
@@ -0,0 +1,184 @@
...
+    actions   = ["servicediscovery:DiscoverInstances"]
+    resources = ["*"]
+  }
+
+  statement {
+    sid       = "logs"
+    effect    = "Allow"
+    actions   = ["logs:CreateLogGroup"]
+    resources = ["*"]
+  }
```
ghost, line 146: Read up on these permissions. Ensure we actually need `["*"]` resources.

modules/teamcity/agents.tf
```diff
@@ -0,0 +1,184 @@
...
+  #checkov:skip=CKV_AWS_158: CW Log Group does not need to be encrypted
+  role       = aws_iam_role.teamcity_agent_task_execution_role.name
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"
+}
+
+resource "aws_cloudwatch_log_group" "teamcity_agent" {
+  #checkov:skip=CKV_AWS_158: CW Log Group does not need to be encrypted
+  #checkov:skip=CKV_AWS_338: CW Log Group does not need to be retained for 1 year
+  name              = "/ecs/teamcity-agent"
+  retention_in_days = 7
```
ghost: This should be available as a variable. Agent log retention should be customizable.

modules/teamcity/main.tf
ghost: More clarification on why the database is breaking!

We should figure out a longer term solution. Or link out to an issue / other documentation outlining this phenomenon.

modules/teamcity/main.tf
```diff
@@ -90,14 +102,19 @@ resource "aws_ecs_task_definition" "teamcity_task_definition" {
...
+  deployment_minimum_healthy_percent = 0
+  deployment_maximum_percent         = 100
+
   wait_for_steady_state = false #TODO: make this configurable
```
ghost, line 123: We might want to make this depend on the `debug` variable. I'm not sure what we do elsewhere...

henrykie (dismissed):
Looks good. Left a few comments on next steps / optional changes. Should those changes not be included in this PR I would like to see them defined in issues.

Also, can you create an issue to scope migrating horde to use service connect or cloudmap for agent -> service communication? The difference there is the agents for Horde are EC2 instances instead of another ECS service.

modules/teamcity/main.tf
ayatanb: The databases were breaking because the tasks would cycle out, meaning that when one was terminating, another one would be booting up. At this time, the Teamcity server thinks that the database is assigned to 2 different servers and you'd be forced to do a completely new deployment because the server won't work. This temp solution works because it makes sure the task is completely off before another one is booted up, and I don't know an alternative yet hahaha. Any ideas?

henrykie (approved):

PR #659, opened 2025-07-16
fix(perforce): fix typo in generating p4 auth service url

dislogical:
## Summary
Fixes a typo causing the P4 module to fail if `p4_server_config.auth_service_url` is not provided and `create_route53_private_hosted_zone` is false.

### Changes

Because module.p4_server is an array now (`count = 0 ? 1`), it must be indexed when read.

## Checklist

* [X] Changes are documented

<details>
No it's a fixing change.
</details>

## Acknowledgment

### DIFF
--- a/modules/perforce/main.tf
+++ b/modules/perforce/main.tf
@@ -17,7 +17,7 @@ module "p4_server" {
       (
         var.create_route53_private_hosted_zone ?
         "auth.${aws_route53_zone.perforce_private_hosted_zone[0].name}" :
-        module.p4_auth.alb_dns_name
+        module.p4_auth[0].alb_dns_name
       ) :
       null
     )


modules/perforce/main.tf
```diff
@@ -17,7 +17,7 @@ module "p4_server" {
       (
         var.create_route53_private_hosted_zone ?
         "auth.${aws_route53_zone.perforce_private_hosted_zone[0].name}" :
-        module.p4_auth.alb_dns_name
+        module.p4_auth[0].alb_dns_name
```
erikyb, line 20: Shouldn't the same be done in `depends_on` later down?
```
depends_on                     = [module.p4_auth]
```

novekm: nah, the count is only necessary here because we are trying to reference an attribute on `p4_auth` and evaluate it as part of a conditional expressions. This value is conditionally created (count 0 or 1) so we need to explicitly define which element in the array we want, which would be 0 since there is only two possible values in this context, 0 (true/create) or 1 (false/do not create). 

I believe for the `depends_on` we can just reference the submodule as is, because we aren't targeting any attributes within it, instead the module itself. So technically it's the somewhat like `p4_auth[*]` in this context

dislogical:
Looks like the tests failed checking out the repo, did I do something wrong opening the PR?

https://github.com/aws-games/cloud-game-development-toolkit/actions/runs/16332069577/job/46136719536#step:6:66

novekm:
Nah not you, pipeline test needs a quick update. I believe you're working off of a fork right? #660 should resolve. That will need to be merged in, then I can re-run tests on your PR cc @henrykie

novekm:
hmm, let me dive into the GH action a bit more - it seems it still can't find the branch named `fix/p4`

Issue #648, opened 2025-07-07
Bug: p4 server vs client file size issue


cw255:
### Expected Behaviour

Expect file sizes to match between client OS and p4 server on submit, allowing submits to complete within server disk space limits

### Current Behaviour

1gb submit comprised of 10500 submit fails

`The filesystem 'blah' has only 98.6G free, but the command estimates it needs at least 102.5G available.`

1gb submit comprised 10000 files succeeds 

Original submit was UE5.6 but seems unrelated to actual source data

### Possible Solution

Update SDP package (once perforce support / can repro and engineer a fix - it has been reported)

### Steps to Reproduce

Create instance with packer as per instructions (I did observe that the device mapping did get muddled somewhere but I can't see how that would result in odd behaviour of the application)

/dev/nvme3n1           128G  2.6G  126G   2% /hxlogs
/dev/nvme2n1            32G  453M   32G   2% /hxmetadata
/dev/nvme1n1           100G  1.3G   99G   2% /hxdepots (I've grown this from 32GB(?), however I think in template its supposed to be 128GB assigned to logs)

Deploy create-resources-complete perforce example with modifications to AWS region and profile only

### Cloud Game Development Toolkit version

latest

cw255:
Confirmed perforce bug between client and server versions

2025.1 client

@pv@ 11 @db.working@ @//bruno_ws/depot/tst@ @//depot/tst@ @bruno_ws@ @bruno@ 0 1 0 0 0 0 0 0 00000000000000000000000000000000 **4** 0 0 0 @@ 0 
 
If you use a 2024.2 or earlier client (on a 2025.1 P4D) a fixed size is inserted into the DB:
 
@pv@ 11 @db.working@ @//bruno_ws/depot/tst@ @//depot/tst@ @bruno_ws@ @bruno@ 0 1 0 0 0 0 0 0 00000000000000000000000000000000 **10485760** 0 0 0 @@ 0

Only seems to apply to new deployments from SDP, upgrading existing server to that version doesn't seem to produce same results.

PR #670, opened 2025-07-24
fix(p4)!: correct volume attachment mapping

dislogical:
**Issue number:** #648 

## Summary

Previously, due to a typo, the logs and depot mappings were swapped. This would result in a 128gb logs volume and a 32gb depot volume, which is undesirable.

### Changes

Use the provided variable names for volume mounts instead of hard-codes (which were incorrect).

## Checklist

* [X] Changes are documented

<details>
Yes, this is a breaking change.
</details>

## Acknowledgment

### DIFF
--- a/modules/perforce/modules/p4-server/main.tf
+++ b/modules/perforce/modules/p4-server/main.tf
@@ -24,8 +24,10 @@ resource "awscc_secretsmanager_secret" "super_user_username" {
 # Perforce P4 Server Instance
 ##########################################
 locals {
-  is_fsxn  = var.storage_type == "FSxN"
   is_iscsi = var.protocol == "ISCSI"
+  is_fsxn  = var.storage_type == "FSxN"
+  is_ebs   = var.storage_type == "EBS"
+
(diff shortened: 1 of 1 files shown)


gabebatista (approved):
lgtm

runreal-warman:
+1 Ran into this issue with the volumes swapped. 

On Nitro based (Graviton) instances the EBS volumes are mounted as `/dev/nvme[0-26]n1` so I believe we need to update the detection in `p4_configure.sh` to account for that as `dev/sd*` ordering is not guaranteed. 

From docs: `The block device driver can assign NVMe device names in a different order than you specified for the volumes in the block device mapping.`

- [nvme ebs docs](https://docs.aws.amazon.com/ebs/latest/userguide/nvme-ebs-volumes.html)
- [attaching nvme volumes](https://support.hashicorp.com/hc/en-us/articles/360058873074-How-to-correctly-handle-attach-EBS-backed-volumes-and-NVMe-on-Linux-instances)

runreal-warman:
Tested this fix and confirming it worked on NVMe instance 👍

PR #694, opened 2025-08-06
Gizzigg/vdi-tf VDI terraform module 

gizzigg:
**Issue number: 694**

### Changes

Added comprehensive VDI module with Active Directory integration and multi-instance deployment capabilities. Includes AD credential functionality and multi-instance deployment.

### User experience

Before: No VDI infrastructure available - users would need to manually configure virtual desktop instances, AD integration, security groups, and networking.

After: Users can deploy production-ready VDI workstations with a single Terraform module call. Includes optional automated AD domain joining, pre-configured security groups, encrypted storage, monitoring, and complete example configurations. Credentials are automatically generated and stored in AWS Secrets Manager for secure access.

## Checklist

* [x] Changes are documented

<details>
no
</details>

## Acknowledgment

### DIFF
--- a/assets/packer/virtual-workstations/README.md
+++ b/assets/packer/virtual-workstations/README.md
@@ -39,9 +39,7 @@ assets/packer/virtual-workstations/
 │   ├── sysprep.ps1               # EC2Launch configuration
 │   └── userdata.ps1              # Packer WinRM setup
 ├── lightweight/               # Base VDI AMI
-├── ue-gamedev/               # Unreal Engine development AMI
-├── artists/                  # Creative tools AMI (planned)
-└── old/                      # Previous templates (archived)
+└── ue-gamedev/               # Unreal Engine development AMI
 ```
 
 ## Prerequisites
@@ -110,19 +108,10 @@ cd assets/packer/virtual-workstations/ue-gamedev/
 packer build windows-server-2025-ue-gamedev.pkr.hcl
 ```
 
-**Includes:** Lightweight base + Visual Studio 2022 + Unreal Engine 5.3
+**Includes:** Lightweight base + Visual Studio 2022 + Epic Games Launcher (UE requires manual install)
 **Build Time:** ~45 minutes
 
-### Artists AMI (Planned)
-**Best for:** Creative workflows
 
-```bash
-# Coming soon
-cd assets/packer/virtual-workstations/artists/
-packer build windows-server-2025-artists.pkr.hcl
-```
-
-**Will Include:** Lightweight base + Blender + Maya + Creative Suite
 
 ## Shared Infrastructure
 
--- a/assets/packer/virtual-workstations/lightweight/README.md
+++ b/assets/packer/virtual-workstations/lightweight/README.md
@@ -42,20 +42,56 @@ This Packer template creates a lightweight Windows Server 2025 AMI optimized for
 - **Instance Type:** g4dn.2xlarge (for GPU driver testing)
 - **Storage:** 80GB root volume
 
-## Usage with VDI Module
+## Using the AMI
 
-Use the resulting AMI with the VDI Terraform module for runtime software customization:
+### Option 1: AWS Console (Manual Deployment)
+
+1. **Launch Instance:**
+   - Go to EC2 Console → Launch Instance
+   - Search for your AMI ID (from Packer build output)
+   - Select GPU instance type: `g4dn.xlarge` or larger
+   - Configure security group to allow RDP (port 3389) or DCV (port 8443)
+
+2. **Connect via DCV:**
+   - Install DCV Client: https://download.nice-dcv.com/
+   - Connect to: `https://<instance-ip>:8443`
+   - Login with Administrator account
+
+3. **Install Additional Software:**
+   - Use Chocolatey: `choco install vscode`
+   - Use PowerShell: Install modules as needed
+   - Manual installers: Download and install directly
+
+### Option 2: Terraform (Infrastructure as Code)
 
 ```hcl
-templates = {
-  "developer" = {
-    instance_type = "g4dn.xlarge"
-    software_packages = [
-      "chocolatey",
-      "git",
-      "visual-studio-2022",
-      "unreal-engine-5.3"
-    ]
+resource "aws_instance" "workstation" {
+  ami           = "ami-0123456789abcdef0"  # Your AMI ID
+  instance_type = "g4dn.xlarge"
+  subnet_id     = var.subnet_id
+
+  vpc_security_group_ids = [aws_security_group.workstation.id]
+
+  tags = {
+    Name = "Developer Workstation"
+  }
+}
+```
+
+### Option 3: VDI Module (Advanced Automation)
+
+For multi-user environments with automated user management:
+
+```hcl
+module "vdi" {
+  source = "path/to/vdi/module"
+
+  presets = {
+    "developer" = {
+      ami           = "ami-0123456789abcdef0"  # Your AMI ID
+      instance_type = "g4dn.xlarge"
+      software_packages = ["vscode", "git"]
+    }
   }
 }
 ```
@@ -63,8 +99,7 @@ templates = {
 ## Alternative AMIs
 
 For faster boot times with pre-installed software, consider:
-- **[UE GameDev AMI](../ue-gamedev/)** - Full Unreal Engine development stack
-- **[Artists AMI](../artists/)** - Creative tools and applications
+- **[UE GameDev AMI](../ue-gamedev/)** - Visual Studio 2022 + Epic Games Launcher (UE requires manual install)
 
 ## Troubleshooting
 
--- a/assets/packer/virtual-workstations/lightweight/windows-server-2025-lightweight.pkr.hcl
+++ b/assets/packer/virtual-workstations/lightweight/windows-server-2025-lightweight.pkr.hcl
@@ -56,7 +56,7 @@ variable "root_volume_size" {
 # Users should not modify this - it aligns with toolkit releases
 locals {
   timestamp = regex_replace(timestamp(), "[- TZ:]", "")
-  version = "v1.1.5"  # Update this for each toolkit release
+  version = "v1.1.6"  # Update this for each toolkit release
   ami_name = "${var.ami_prefix}-${local.timestamp}"
 }
 
--- a/assets/packer/virtual-workstations/ue-gamedev/windows-server-2025-ue-gamedev.pkr.hcl
+++ b/assets/packer/virtual-workstations/ue-gamedev/windows-server-2025-ue-gamedev.pkr.hcl
@@ -56,7 +56,7 @@ variable "root_volume_size" {
 # Users should not modify this - it aligns with toolkit releases
 locals {
   timestamp = regex_replace(timestamp(), "[- TZ:]", "")
-  version = "v1.1.5"  # Update this for each toolkit release
+  version = "v1.1.6"  # Update this for each toolkit release
   ami_name = "${var.ami_prefix}-${local.timestamp}"
 }
 


modules/vdi/examples/simple/directory.tf
```diff
@@ -0,0 +1,77 @@
...
+# Security Group for Simple AD
+resource "aws_security_group" "simple_ad_sg" {
+  count       = var.enable_simple_ad ? 1 : 0
+  name_prefix = "${local.name_prefix}-simple-ad-"
+  vpc_id      = aws_vpc.vdi_vpc.id
+  description = "Security group for Simple AD"
+
+  # Allow all traffic within VPC for AD communication
+  ingress {
+    from_port   = 0
```
IAmSkweetis: Consider reducing the scope of inbound rules, even in the VPC.

My suggestion would be to scope this down to just the ports required for AD and with a source of the domain controller's IP.

For reference: https://learn.microsoft.com/en-us/troubleshoot/windows-server/active-directory/config-firewall-for-ad-domains-and-trusts

modules/vdi/sg.tf
```diff
@@ -0,0 +1,84 @@
...
+  name_prefix = "${var.project_prefix}-${var.name}-vdi-"
+  vpc_id      = var.vpc_id
+  description = "Security group for VDI instances"
+
+  # RDP access from user public IP
+  ingress {
+    from_port   = 3389
+    to_port     = 3389
+    protocol    = "tcp"
+    cidr_blocks = ["${chomp(data.http.user_public_ip.response_body)}/32"]
```
IAmSkweetis: Consider adding in an optional input variable for an override subnet. If the user is behind a firewall that uses a NAT pool, the response back on the data resource will vary. 

For example, consider a user in an office that has a NAT pool of `192.0.2.0/24`. The user may get an IP of `192.0.2.5` on the initial run of Terraform. But when they attempt to use DCV to access the instance, the user might grab `192.0.2.134` on their next session through the firewall.

modules/vdi/variables.tf
```diff
@@ -81,14 +79,21 @@ variable "create_key_pair" {
 
 variable "admin_password" {
```
IAmSkweetis: Consider moving this into proper secret management. While `sensitive=true` will prevent the password from being printed to stdout/stderr, there is no provided encryption within the state file.

modules/vdi/README.md
```diff
@@ -262,7 +289,7 @@ Replace `<secret-id>` with the value of the `secrets_manager_secret_id` output.
 
 ## TODO
 
-- [ ] Replace VPC placeholder with actual VPC reference when VPC module is identified
+- [x] Add support for creating VPC with public and private subnets
 - [ ] Add support for Auto Scaling Groups for multiple VDI instances
 - [ ] Add CloudWatch monitoring and alerting?
 - [ ] Add backup and snapshot policies?
```
awsbilal: Please include a cost estimate in the README as well, so customers arent surprised by cost.

awsbilal: Per Kevon: Let's add Terraform Test to this module

modules/vdi/variables.tf
novekm: Sounds like a great use case for Ephemeral Resources? https://developer.hashicorp.com/terraform/language/resources/ephemeral :)

modules/vdi/sg.tf
```diff
@@ -0,0 +1,84 @@
+# Security group for the VDI instance
+resource "aws_security_group" "vdi_sg" {
+  name_prefix = "${var.project_prefix}-${var.name}-vdi-"
+  vpc_id      = var.vpc_id
+  description = "Security group for VDI instances"
+
+  # RDP access from user public IP
+  ingress {
```
ayatanb: aren't inline arguments for ingress and egress discouraged by Terraform? In our other modules, we usually use the [aws_vpc_security_group_egress_rule](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpc_security_group_egress_rule) (and similar note for ingress). I might be wrong, but I see a note on the terraform provider page:

NOTE:
Avoid using the ingress and egress arguments of the aws_security_group resource to configure in-line rules, as they struggle with managing multiple CIDR blocks, and, due to the historical lack of unique IDs, tags and descriptions. To avoid these problems, use the current best practice of the [aws_vpc_security_group_egress_rule](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpc_security_group_egress_rule) and [aws_vpc_security_group_ingress_rule](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpc_security_group_ingress_rule) resources with one CIDR block per rule.

modules/vdi/sg.tf
joshral7459: I removed the function of auto IP lookup. I think for now having a admin input public IPs manually is the most general approach. can address edge cases as they come.

modules/vdi/variables.tf
joshral7459: Refactored passwords completely. All stored in secrets or need a keypair to retrieve.

modules/vdi/README.md
joshral7459: done

joshral7459: done

modules/vdi/sg.tf
joshral7459: fixed

modules/vdi/sg.tf
kylesomers: Discuss with @novekm. He implemented something similar in the Perforce module [here](https://github.com/aws-games/cloud-game-development-toolkit/blob/0d3682c28df27afd0f476c7eb8bb2c4c6eaa2bda/modules/perforce/examples/create-resources-complete/security.tf#L11). I'd like us to maintain consistency in our approach to retrieving end user IP address to provide best user experience across modules.

kylesomers: Also [here](https://github.com/aws-games/cloud-game-development-toolkit/blob/0d3682c28df27afd0f476c7eb8bb2c4c6eaa2bda/modules/unity/accelerator/examples/simple/main.tf#L10C6-L10C40)

awsbilal (dismissed):
LGTM

modules/vdi/sg.tf
novekm: resolved in refactor

assets/packer/virtual-workstations/lightweight/README.md
ghost, line 1: What is the purpose of two separate templates? We can control the additional installation of UE tooling (which it seems is just visual studio and the epic games launcher) using variables.

assets/packer/virtual-workstations/lightweight/README.md
```diff
@@ -0,0 +1,73 @@
...
+    ]
+  }
+}
+```
+
+## Alternative AMIs
+
+For faster boot times with pre-installed software, consider:
+- **[UE GameDev AMI](../ue-gamedev/)** - Full Unreal Engine development stack
+- **[Artists AMI](../artists/)** - Creative tools and applications
```
ghost: This doesn't exist.

assets/packer/virtual-workstations/lightweight/README.md
```diff
@@ -0,0 +1,73 @@
...
+   packer build -var-file="variables.pkrvars.hcl" windows-server-2025-lightweight.pkr.hcl
+   ```
+
+## Build Time
+
+- **Estimated:** 20-30 minutes
+- **Instance Type:** g4dn.2xlarge (for GPU driver testing)
+- **Storage:** 80GB root volume
+
+## Usage with VDI Module
```
ghost: Why is this here? This AMI should be a discrete tool that is usable without the module. We should provide guidance on how you can leverage this AMI via the console as well.

assets/packer/virtual-workstations/ue-gamedev/unreal_development_stack.ps1
ghost, line 1: Just make this an additional provisioner that runs based on a variable with a single packer template...

assets/packer/virtual-workstations/README.md
```diff
@@ -0,0 +1,199 @@
...
+```
+assets/packer/virtual-workstations/
+├── shared/                    # REQUIRED - Base infrastructure scripts
+│   ├── base_infrastructure.ps1    # NVIDIA + DCV + AWS tools + dev tools
+│   ├── sysprep.ps1               # EC2Launch configuration
+│   └── userdata.ps1              # Packer WinRM setup
+├── lightweight/               # Base VDI AMI
+├── ue-gamedev/               # Unreal Engine development AMI
+├── artists/                  # Creative tools AMI (planned)
+└── old/                      # Previous templates (archived)
```
ghost: inaccurate.

assets/packer/virtual-workstations/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+```bash
+# Clone the entire repository
+git clone https://github.com/aws-games/cloud-game-development-toolkit.git
+cd cloud-game-development-toolkit
+
+# Verify structure exists
+ls assets/packer/virtual-workstations/shared/
+# Should show: base_infrastructure.ps1  sysprep.ps1  userdata.ps1
+```
```
ghost, line 57: I really feel like this is unnecessary. We're in the repo already. We're trying to provide a single packer template and some provisioner scripts. Insinuating that we need everything else in the Toolkit is inaccurate.

assets/packer/virtual-workstations/README.md
```diff
@@ -0,0 +1,199 @@
...
+```bash
+# Create variables file (optional)
+cp variables.pkrvars.hcl.example variables.pkrvars.hcl
+
+# Edit variables.pkrvars.hcl
+instance_type = "g4dn.4xlarge"  # Must be GPU-enabled
+
+# Build with custom variables
+packer build -var-file="variables.pkrvars.hcl" windows-server-2025-lightweight.pkr.hcl
+```
```
ghost, line 82: Why not just link to the packer documentation on passing variables? We're re-inventing existing docs.

assets/packer/virtual-workstations/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Available Templates
+
+### Lightweight AMI
+**Best for:** Runtime software customization via VDI Terraform module
```
ghost, line 87: I'm not sure on why we would do this. I think just building multiple AMIs makes way more sense. Why are we offloading software installation to the TF module?

ghost: To clarify, I don't mean multiple templates. I mean running the following:

`packer build -var "unreal=True" my_template.pkr.hcl`

and then running

`packer build -var "blender=True" my_template.pkr.hcl`

and then provisioning the VDI module with both if you want, or using the built AMIs directly in the console. We're treating it like a pre-req to the module, but we want these things decoupled.

assets/packer/virtual-workstations/README.md
ghost: https://developer.hashicorp.com/packer/guides/hcl/variables#assigning-variables

assets/packer/virtual-workstations/README.md
```diff
@@ -0,0 +1,199 @@
...
+## Template Dependencies
+
+**Each template references shared scripts:**
+- `../shared/base_infrastructure.ps1` - Common infrastructure setup
+- `../shared/sysprep.ps1` - EC2Launch configuration
+- `../shared/userdata.ps1` - Packer WinRM connectivity
+
+**This is why the complete directory structure is required.**
+
+## Usage with VDI Module
```
ghost, line 137: Again, this documentation belongs in the module. We can link to it, but the AMIs can also be used via the console, or pulumi, or any other number of tools. We don't want to establish a strict dependency here.

modules/jenkins/README.md
ghost, line 1: these should really be separate PRs.

modules/perforce/README.md
ghost, line 1: should be separate PR.

modules/vdi/examples/private-connectivity/main.tf
```diff
@@ -0,0 +1,188 @@
...
+  region         = data.aws_region.current.id
+  environment    = "dev"
+  vpc_id         = aws_vpc.vdi_vpc.id
+
+  # Create Client VPN infrastructure
+  create_client_vpn = true
+
+  # Client VPN configuration - avoid CIDR overlap with local network
+  client_vpn_config = {
+    client_cidr_block = "10.100.0.0/16" # Avoid overlapping with local networks, especially 192.168.x.x/24 home networks
```
ghost, line 21: I do feel like this should be external to the module. The module really should handle instance provisioning, user management and authentication, shared storage / mounting, and security group / network access to to shared services. For example:

We install Perforce on the instances at the AMI level. However, configuring the client to point at the correct P4 Server could be done at runtime by setting an environment variable on deployment (P4PORT).

If Horde is present we could automate the downloading of the Horde Agent from the Horde server and auto enroll and register the VDI instance. Same goes for DDC theoretically.

modules/vdi/examples/private-connectivity/main.tf
```diff
@@ -0,0 +1,188 @@
...
+  }
```
ghost, line 22: I do feel like this should be external to the module. The module really should handle instance provisioning, user management and authentication, shared storage / mounting, and security group / network access to to shared services. For example:

We install Perforce on the instances at the AMI level. However, configuring the client to point at the correct P4 Server could be done at runtime by setting an environment variable on deployment (P4PORT).

If Horde is present we could automate the downloading of the Horde Agent from the Horde server and auto enroll and register the VDI instance. Same goes for DDC theoretically.

modules/vdi/examples/private-connectivity/main.tf
```diff
@@ -0,0 +1,188 @@
...
+          windows_drive = "D:"
+          iops          = 3000
+          encrypted     = true
+        }
+      }
+      # Full DevOps toolchain
+      software_packages = [
+        "vscode", "terraform", "kubernetes-cli", "docker-desktop",
+        "postman", "checkov", "7zip"
+      ]
```
ghost, line 76: Okay, I can see the value here. But I still think that we could include this variable when building the AMI. whats the benefit of waiting to do it here?

modules/vdi/examples/private-connectivity/main.tf
```diff
@@ -0,0 +1,188 @@
...
+  # Create Client VPN infrastructure
+  create_client_vpn = true
+
+  # Client VPN configuration - avoid CIDR overlap with local network
+  client_vpn_config = {
+    client_cidr_block = "10.100.0.0/16" # Avoid overlapping with local networks, especially 192.168.x.x/24 home networks
+  }
+
+  # 1. PRESETS - Configuration blueprints with different AMIs for different roles
+  presets = {
```
ghost, line 25: So this maps to individual "fleets" of instances? I can set N presets? Where do I define how many instances of each preset I want?

modules/vdi/examples/private-connectivity/main.tf
```diff
@@ -0,0 +1,188 @@
...
+      use_client_vpn = true            # VPN access - gets .ovpn file
+    }
+    # Junior Developer - standard user, VPN access for learning
+    "boruto-uzumaki" = {
+      given_name     = "Boruto"
+      family_name    = "Uzumaki"
+      email          = "boruto@example.com"
+      type           = "user" # Windows Users group - limited privileges
+      use_client_vpn = true   # VPN access - gets .ovpn file
+    }
```
ghost, line 170: Where are these stored?

modules/vdi/client-vpn.tf
```diff
@@ -0,0 +1,380 @@
...
+
+##########################################
+# Client VPN Server Certificate (Self-signed)
+##########################################
+resource "tls_private_key" "client_vpn_server" {
+  count = local.enable_client_vpn ? 1 : 0
+
+  algorithm = "RSA"
+  rsa_bits  = 2048
+}
```
ghost, line 23: so we're doing cert based as opposed to directory based auth for both the instances and the client VPN? We can't use directory service for client VPN access _or_ instance auth?

https://docs.aws.amazon.com/directoryservice/latest/admin-guide/what_is.html

henrykie (changes requested):
@novekm @awsbilal @joshral7459 @gizzigg I left a lot of comments, but the main thing I want to see before approval is slight tweaks to the documentation (fix inaccuracies) and maybe re-evaluate the multi packer template approach. I don't think thats necessary. I also dont think package installation is a "runtime" decision that we need to offer via the module. I think we can achieve the same "flexibility" through a variable in Packer, and thats more flexible. What if I don't want to use the VDI module but I want to include other packages in my AMI? I need to modify the packer template to do that today. Just move the variable over.

My concerns with the module structure (active directory, directory service, client VPN) don't need to be resolved for approval. Does that make sense?

Thanks!

assets/packer/virtual-workstations/lightweight/README.md
novekm: My thought was in the future we may add additional Packer templates beyond game dev only (e.g. art focused with maya, blender, audio focused with Wwise, FMOD, Pro Tools, Ableton etc). To ensure consistency I thought a shared template could be used for general VDI stuff (NVIDIA drivers, Perforce, Git, etc) so if we need to make updates we can do it in one place. Would those also be variables?

modules/vdi/examples/private-connectivity/main.tf
novekm: Secrets manager alongside the windows password for users. Then relevant values are read from secrets manager when creating the user account(s) via SSM

modules/vdi/client-vpn.tf
novekm: we synced on this and decided AD should be out of scope for v1 of the module - local user management is all that is supported out of the box right now, but users could integrate with AD on their own for the instances that are provisioned. Really we're looking to IAM IdC for SSO for v2 (AD could be one of these). Part of the challenge was with AD there's no way to actually create/manage users/groups in Terraform like you can do with the [aws_iam_identitystore_user](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/identitystore_user) and [aws_iam_identitystore_group resources](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/identitystore_group).

This means that there's no real mapping of the users you'd define in the module and the actual AD users/groups you'd want to create. The only real relevant resource is used to create the directory service directory itself ([aws_directoryservice_directory](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/directory_service_directory)) which imo doesn't really provide much value on its own. Users would still need to create their own users and groups elsewhere (manually, additional scripting outside the module etc) which kinda devalued having this in the module. We looked into the [AD provider](https://github.com/hashicorp/terraform-provider-ad) to hopefully enable this functionality, but its been archived.

modules/vdi/examples/private-connectivity/main.tf
novekm: Customization and speed of use - at times it takes close to an hour to build the AMI with UE stuff. The idea is that this is much quicker for adhoc stuff, and allows for customization on user-by-user basis without having to rebuild a ton of AMIs or run your own manual SSM docs/scripting. Also for post-deployment functionality - if a user forgets to bake something into the AMI they'd have to rebuild a new one (up to an hour) and then deal with the deletion and recreation of the instance because of the AMI changing (and the related EBS stuff to ensure no data loss).

modules/vdi/examples/private-connectivity/main.tf
novekm: so presets just map to the configuration you want for N instances, they don't map to entire fleets. You assign the preset on each instance you want to use it. Perhaps the naming of this variable could be more clear, but presets would include things like:
- ami
- instance config
- volume mapping
- additional software (if any)
- etc

the idea is that there are 3 main ways to configure the instances:
1. using presets based on job role like `gamedev`, `devops`, `art` etc
2. using presets and also customizing if needed with the `software_packages` attribute (more for adhoc stuff or if you forget to bake into the ami)
3. direct instance-by-instance configuration (with or without the `software_packages` defined)

Each instance then can use a preset if desired (options 1 + 2), or not (option 3). If a preset is desired to be used for the instance, it can be set using the `preset_key` attribute
e.g.
```hcl
# define presets that you want available to be applied to workstations (defined on each workstation desired)
presets = {
    # Game developer workstation with pre-built UE + Visual Studio
    ue-gamedev-workstations = {
      instance_type = "g4dn.4xlarge"  
      #...additional config
    },
    devops_workstations = {
    # devops stuff
    }
}

# each workstation can use a preset if they want, or not
workstations = {
    # Pattern 1: Pure template-based (Naruto - Game Developer)
    "vdi-001" = {
      preset_key          = "ue-gamedev-workstation" # Uses template exactly as defined (KEY MATCHES PRESET DEFINED ABOVE)
      assigned_user       = "naruto-uzumaki"         # User assigned to this workstation
      subnet_id           = aws_subnet.vdi_subnet.id
      security_groups     = [aws_security_group.vdi_sg.id]
      allowed_cidr_blocks = ["${chomp(data.http.my_ip.response_body)}/32"]
    }
}
```

assets/packer/virtual-workstations/ue-gamedev/unreal_development_stack.ps1
novekm: I see where you're coming from with the conditional provisioners - it's definitely an elegant solution. However, given all the creative software we may plan to support (Unreal for now, but maybe Maya, Blender, Unity, Pro Tools, Davinci Resolve, etc.), I think separate templates would serve users better in the long run. A single template might get pretty complex to maintain as we add more software options. Building off a base template could give us cleaner, more focused configurations while still keeping the core stuff consistent. Users could just grab the template they need without having to navigate through all the different options. What do you think?

assets/packer/virtual-workstations/README.md
novekm: removed the unused directories from the readme

Issue #429, opened 2025-01-06
Deploy floating Unity license server for build farm


henrykie:
### Unity Floating License Server

The Cloud Game Development Toolkit should provide a module for provisioning a floating license server for Unity build machines. [Here is the documentation.](https://docs.unity.com/licensing/en-us/manual)

gabebatista:
It might be worth opening an RFC for Unity support in general. It would be good for us to get a fly-by overview of what we could do for Unity (since a lot of the services they offer are SaaS based).

henrykie:
@gabebatista @danielw-aws please see[ this discussion](https://github.com/aws-games/cloud-game-development-toolkit/discussions/431) for extended discussion on Unity support in the Toolkit. @gabebatista - I leave it to you to decide if we should keep this issue open, or close it for now until we flesh out more of the conversation.

henrykie:
@danielw-aws let's flesh this out a bit more before next meeting.

henrykie:
FYI there is an existing CDK based guidance on this: https://github.com/aws-samples/unity-build-server-with-aws-cdk

gabebatista:
Closing: Implemented. The `modules/unity/floating-license-server/` module is complete with ECS deployment, ALB, variables, outputs, and examples.

PR #745, opened 2025-09-24
feat(unity): deploy floating licensing server (#429)

erikyb:
**Issue number:**
#429 

## Summary

This PR introduces the Unity Licensing Server, a server that manages a pool of Unity Editor floating licenses for organizations, allowing clients to check out and release licenses as they start and close the Unity Editor.

### Changes

The deployment uses an Amazon EC2 instance to run the Unity Licensing Server. An Amazon Elastic Network Interface (ENI) is attached to the instance to provide a static private (and public if configured) IP to separate network settings from the compute instance, as the Unity Licensing Server binds the license to the machine's identity. The Licensing Server provides an admin dashboard. The username is admin and cannot be changed. The password is generated securely by the module and stored in AWS Secrets Manager. An Amazon S3 bucket is created and mounted to the instance. This bucket is used to provide the user with files created by the Unity Licensing Server, as well as to provide files to the Licensing Server that will need to import in later steps.

### User experience

Installing and configuring the Unity Licensing Server is manual, multi-step process. This module simplifies the steps and interactions needed to successfully start the server by piping commands to set up the server on behalf of the user, mounting an S3 bucket for input and output of files need for and produced by the server, and configuring background processes that detect the file drops and run the licensing import on behalf of the user. The module also provides the necessary configurations to set up the Licensing Server as an EC2 instance in a public subnet for simplified, public-facing deployment, a private subnet for internal usage only, and a private subnet behind an Application Load Balancer for a more secure and robust public deployment.

## Checklist

* [X] Changes are documented

<details>
No.

</details>

## Acknowledgment


modules/unity/floating-license-server/examples/simple/main.tf
```diff
@@ -0,0 +1,44 @@
+module "unity_license_server" {
+  source                         = "../../"
+  unity_license_server_file_path = "/path/to/your/server/file/Unity.Licensing.Server.linux-x64-v2.1.0.e3f6bc2.zip"
```
ghost: If this is a required variable then make it a required variable at the example level. When a user runs `terraform apply` against the simple example configuration they'll be prompted to provide this path. Seem reasonable?

ghost: You can also use some of Terraform value referencing syntax to get creative with path definitions. [Check out this documentation.](https://developer.hashicorp.com/terraform/language/expressions/references#filesystem-and-workspace-info)

modules/unity/floating-license-server/examples/simple/main.tf
```diff
@@ -0,0 +1,44 @@
...
+  enable_alb_deletion_protection = false
+
+  depends_on = [
+    aws_nat_gateway.nat_gateway,
+    aws_route.private_rt_nat_gateway,
+    aws_eip.nat_gateway_eip,
+    aws_route_table.private_rt,
+    aws_route_table.public_rt,
+    aws_route_table_association.private_rt_asso,
+    aws_route_table_association.public_rt_asso
```
ghost, line 19: Did you find these were all necessary to successfully deploy?

modules/unity/floating-license-server/scripts/license_daemon.sh
```diff
@@ -0,0 +1,229 @@
+#!/bin/bash
+
+echo "[$(date '+%Y-%m-%d %H:%M:%S')] Setting up license watch daemon..."
```
ghost: consider incorporating this prefix into a log function that you call regularly. [Here is an example from the `p4_configure.sh` script.](https://github.com/aws-games/cloud-game-development-toolkit/blob/c386148e6aff0aef0ff033321a46077e0b7fbd6e/assets/packer/perforce/p4-server/p4_configure.sh#L4-L17)

modules/unity/floating-license-server/scripts/license_daemon.sh
```diff
@@ -0,0 +1,229 @@
...
+
+    if [ "$filename" = "${license_server_name}.zip" ]; then
+        echo "Target file detected through inotify"
+        process_license_file
+        if [ $? -eq 0 ]; then
+            exit 0
+        fi
+    fi
+done < <(inotifywait -m -e create,moved_to /mnt/s3)
+WATCHSCRIPT
```
ghost: We're using a script to create a script? Is this because user_data only calls one? I'd pull this script out into another .sh in the same directory just for readability, and then if we need to inject it somewhere we do so using [`file()`](https://developer.hashicorp.com/terraform/language/functions/file)

modules/unity/floating-license-server/scripts/license_daemon.sh
```diff
@@ -0,0 +1,229 @@
...
+User=ubuntu
+Group=unity-licensing-server
+WorkingDirectory=/opt/UnityLicensingServer
+ExecStart=/bin/bash -l -c '/opt/UnityLicensingServer/watch_for_license_upload.sh'
+StandardOutput=journal
+StandardError=journal
+
+[Install]
+WantedBy=multi-user.target
+WATCHSERVICE
```
ghost: same thing here. I'd consider pulling out the service definition into its own file for easier readability.

modules/unity/floating-license-server/scripts/server_setup.sh
```diff
@@ -0,0 +1,122 @@
...
+
+echo "[$(date '+%Y-%m-%d %H:%M:%S')] Installing prerequisites..."
+apt-get update
+apt-get install -y fuse3 s3fs unzip expect
+
+# Install and start SSM Agent
+#echo "[$(date '+%Y-%m-%d %H:%M:%S')] Installing and starting SSM Agent..."
+#snap install amazon-ssm-agent --classic
+#systemctl enable snap.amazon-ssm-agent.amazon-ssm-agent.service
+#systemctl start snap.amazon-ssm-agent.amazon-ssm-agent.service
```
ghost: nit: remove

modules/unity/floating-license-server/scripts/server_setup.sh
```diff
@@ -0,0 +1,122 @@
...
+expect "Enter the index number of the network interface"
+send "2\r"
+expect "Enter server's listening port number"
+send "$serverport\r"
+expect "Create a password"
+send "$password\r"
+expect "Confirm the password"
+send "$password\r"
+expect eof
+EXPECT
```
ghost, line 63: If possible, lets pull this out into a different file for readability. (if its not possible thats okay too)

modules/unity/floating-license-server/scripts/user_data.tpl
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# Send all output to console and log file
+exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
```
ghost, line 4: Ah - I see. You're collecting logs at the user-data level. I still think a log function might make your life easier in the shell scripts even if its just writing to stdout

modules/unity/floating-license-server/outputs.tf
```diff
@@ -0,0 +1,64 @@
+output "eni_id" {
+  description = "Elastic Network ID (ENI) used when binding the Unity Floating License Server."
+  value       = local.eni_id
+}
+
+output "instance_public_ip" {
+  value = aws_instance.unity_license_server.public_ip
+}
```
ghost, line 9: nit: description

modules/unity/floating-license-server/README.md
```diff
@@ -0,0 +1,148 @@
...
+- The server's MAC address
+- The server's operating system
+- The number of processor cores on the server
+- The name of the server
+
+Note that if the Licensing Server needs to be hosted by a new machine, a support ticket will need to be submitted to [Unity Support](https://support.unity.com/hc/en-us/requests/new) to revoke the Licensing Server so a new one can be configured.
+
+The Licensing Server provides an admin dashboard. The username is `admin` and cannot be changed. The password is generated securely by the module and stored in [AWS Secrets Manager](https://aws.amazon.com/secrets-manager/).
+
+An [Amazon S3](https://aws.amazon.com/pm/serv-s3/) bucket is created and mounted to the instance. This bucket will be used to provide the user with files created by the Unity Licensing Server, as well as to provide files to the Licensing Server that will need to import in later steps.
```
ghost: typo: "will need to [be] imported in later steps"

modules/unity/floating-license-server/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+Upon deployment, two created files will be copied to the S3 bucket:
+
+- The services configuration file `services-config.json`, which should be deployed to all client computers intending to use the Licensing Server.
+- The server registration request file `server-registration-request.xml`, which contains machine binding information from the server.
+
+The `server-registration-request.xml` will need to be uploaded to the [Unity ID portal](https://id.unity.com/) (where the Licensing Server executable was downloaded from) to register the server. Once successful, download the licenses zip file. Without renaming the file, upload it to the S3 bucket. A background process will detect the uploaded licenses file, and import them into the Unity Licensing Server, finishing the process.
```
ghost, line 24: so cool that this works. Great work Erik!

modules/unity/floating-license-server/variables.tf
```diff
@@ -0,0 +1,175 @@
...
+
+#############################################################
+# Unity Floating License Server EC2 Instance Configuration
+#############################################################
+
+variable "unity_license_server_instance_ami_id" {
+  type        = string
+  description = "The AMI ID to use in the EC2 instance running the Unity Floating License Server. Defaults to the Ubuntu Server 24.04 LTS (HVM) AMI ID. Note that this option is provided to specify newer versions of the Ubuntu AMI only."
+  default     = "ami-0d1b5a8c13042c939"
+}
```
ghost, line 118: The AMI ID is region specific I believe, which means this will only work in us-west-2 (or wherever you were testing). Instead, we should use a `data` block to look up the Ubuntu AMI that you want to use by default. This can also pull the latest version.

An escape hatch that supports passing an AMI is still a good idea, but if its null we should use the AMI fetched via `data` block.

modules/unity/floating-license-server/variables.tf
```diff
@@ -0,0 +1,175 @@
...
+variable "unity_license_server_instance_ebs_size" {
+  type        = string
+  default     = "20"
+  description = "The size of the EBS volume in GB."
+}
+
+variable "enable_instance_detailed_monitoring" {
+  type        = bool
+  default     = false
+  description = "Enables detailed monitoring for the instance by increasing the frequency of metric collection from 5-minute intervals to 1-minute intervalss in CloudWatch to provide more granular data. Note this will result in increased cost."
```
ghost: nit: typo - "intervalss"

modules/unity/floating-license-server/variables.tf
```diff
@@ -0,0 +1,175 @@
...
+  type        = bool
+  default     = true
+  description = "If true, enables EC2 instance termination protection from AWS APIs and console."
+}
+
+####################################################
+# Unity Floating License Server Configuration
+####################################################
+
+variable "unity_license_server_file_path" {
```
ghost, line 148: This is where you can get creative with the path functions available in Terraform: https://developer.hashicorp.com/terraform/language/expressions/references#filesystem-and-workspace-info

henrykie (changes requested):
Yo! Looks great. A few things included in the review and one general note:

We should include a `tftest` for this module. It will just deploy the `simple` example you've written in an automated way. @novekm has a bunch of good examples in both the Unreal Cloud DDC and Perforce work he's done.

Otherwise, this is a great PR! Hopefully customers find this useful.

modules/unity/floating-license-server/examples/simple/main.tf
erikyb: There were race conditions where the NAT Gateway-related resources would not complete creation before the EC2 and its user data script started executing, and would fail to connect and download dependencies.

modules/unity/floating-license-server/main.tf
```diff
@@ -0,0 +1,412 @@
...
+}
+
+# Data source to check instance status
+data "aws_instance" "unity_license_server" {
+  instance_id = aws_instance.unity_license_server.id
+  depends_on  = [aws_instance.unity_license_server]
+}
+
+# Resource to wait for instance to be ready and user data to complete
+/*
```
erikyb: Missed this legacy/older code I was trying. Will remove.

modules/unity/floating-license-server/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+While the Unity Licensing Server deployment has been simplified and streamlined in this module, it remains a multi-step process requiring manual intervention. A local zip file of the Unity Licensing Server is required when configuring and running the module. It can be downloaded from the `Organizations` section in the [Unity ID portal](https://id.unity.com/). For simplicity and cost optimization, this module requires the Linux version of the executable.
```
gabebatista, line 4: Might be worth calling out exactly why the zip file is required here. Is this a legal-type issue or is there some technical barrier that keeps us from automating the download?

modules/unity/floating-license-server/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+The deployment uses an [Amazon EC2 instance](https://aws.amazon.com/pm/ec2/) to run the Unity Licensing Server. An [Amazon Elastic Network Interface](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html) (ENI) is attached to the instance to provide a static private (and public if configured) IP to separate network settings from the compute instance, as the Unity Licensing Server binds the license to the machine's identity, which includes the following information:
+
+- The server's MAC address
+- The server's operating system
+- The number of processor cores on the server
+- The name of the server
+
+Note that if the Licensing Server needs to be hosted by a new machine, a support ticket will need to be submitted to [Unity Support](https://support.unity.com/hc/en-us/requests/new) to revoke the Licensing Server so a new one can be configured.
```
gabebatista, line 13: This is interesting, how does this affect `tf destroy`?  If someone needs to redeploy their licensing servers they need to create a ticket with Unity?

modules/unity/floating-license-server/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+The Licensing Server provides an admin dashboard. The username is `admin` and cannot be changed. The password is generated securely by the module and stored in [AWS Secrets Manager](https://aws.amazon.com/secrets-manager/).
```
gabebatista, line 15: Is the `admin` username hardcoded or is this just how the license server is programmed?

modules/unity/floating-license-server/README.md
gabebatista: +1

modules/unity/floating-license-server/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+Upon deployment, two created files will be copied to the S3 bucket:
+
+- The services configuration file `services-config.json`, which should be deployed to all client computers intending to use the Licensing Server.
+- The server registration request file `server-registration-request.xml`, which contains machine binding information from the server.
+
+The `server-registration-request.xml` will need to be uploaded to the [Unity ID portal](https://id.unity.com/) (where the Licensing Server executable was downloaded from) to register the server. Once successful, download the licenses zip file. Without renaming the file, upload it to the S3 bucket. A background process will detect the uploaded licenses file, and import them into the Unity Licensing Server, finishing the process.
+
+## Deployment Architectures
+![Unity Accelerator Module Architecture](./assets/media/diagrams/unity-licensing-server-architecture1.png)
```
gabebatista: you used `Unity Accelerator Module Architecture` for the `alt` value attached to the image. You can't really see this just reading through the docs but it would really confuse someone using a screen reader.

modules/unity/floating-license-server/README.md
```diff
@@ -0,0 +1,148 @@
...
+Simple architecture where clients connect directly to an EC2 instance in a public (or private) subnet through a primary ENI with static public/private IPs. The EC2 instance hosts the Unity Licensing Server and connects to AWS Secrets Manager and Amazon S3 (mounted via s3fs).
+
+![Unity Accelerator Module Architecture](./assets/media/diagrams/unity-licensing-server-architecture2.png)
```
gabebatista: `alt` name (see comment above).

modules/unity/floating-license-server/scripts/user_data.tpl
gabebatista, line 4: I'm assuming this is done this way because of how the null_resource `wait_for_user_data` is implemented in `main.tf`?

modules/unity/floating-license-server/main.tf
```diff
@@ -0,0 +1,412 @@
...
+
+  # Specify the network interface as the primary network interface
+  network_interface {
+    network_interface_id  = local.eni_id
+    delete_on_termination = false
+    network_card_index    = 0
+    device_index          = 0
+  }
+
+  user_data = templatefile("${path.module}/scripts/user_data.tpl", {
```
gabebatista, line 273: I wish there was a cleaner way to implement this (I am not saying I have a better way, call it wishful thinking).

modules/unity/floating-license-server/scripts/server_setup.sh
gabebatista, line 63: I wasn't familiar with `expect`, will def be adding that to my repertoire.

gabebatista: +1

gabebatista (approved):
Wow! Fantastic work, I might even go as far as saying that this is our new golden example of what a module should look like. Henry had already commented on all the little things I notice so most of my comments are either reinforcing what he said or asking question so that I can better understand the design. Again, great work and as always thank you for the contribution to the project!

modules/unity/floating-license-server/README.md
erikyb: It is specifically downloaded from your Unity Organizations portal. The Unity Licensing Server is only available for those with an Enterprise subscription, so we don't want to run the risk of providing software to people that do not qualify for it. It won't do them any good regardless as they will not be able to register the server and get the licenses file, but why risk it.

modules/unity/floating-license-server/README.md
erikyb: Yes. The instance has termination protection from the console and API, but not terraform. tf destroy will delete it. I don't think we should prevent users from doing something they want to do, as long as the consequences are clear. Revoking the license is a simple ordeal, you submit a ticket identifying which organization and the name of the licensing server to revoke, and it is usually taken care of in a day. Had to do it a few times through this process!

modules/unity/floating-license-server/README.md
erikyb: It is hard coded and the Unity Licensing Server offers no way to change this, it only gives you the option of setting the password, not the user name.

modules/unity/floating-license-server/README.md
erikyb: Great catch!

modules/unity/floating-license-server/scripts/user_data.tpl
erikyb: Right, it's logging to `/var/log/user-data.log`, output will also appear in `syslog` (which you can see with `journalctl` or in `/var/log/syslog`) and in the EC2 console logs. The `null_resource` is checking for completion by waiting for a ready state and by looking for the completion token `[END_UDS_TKN]` in the console output.

modules/unity/floating-license-server/scripts/license_daemon.sh
erikyb: Done.

modules/unity/floating-license-server/scripts/license_daemon.sh
erikyb: Separated all nested files into individual files and loading them instead.

modules/unity/floating-license-server/scripts/license_daemon.sh
erikyb: Done.

modules/unity/floating-license-server/variables.tf
erikyb: I was unaware of this, thanks for the great suggestion. Researched how to make this happen and got it working.

henrykie (dismissed):
LGTM Erik! Nice work.

gabebatista (approved):
LGTM

modules/unity/floating-license-server/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+While the Unity Licensing Server deployment has been simplified and streamlined in this module, it remains a multi-step process requiring manual intervention. A local zip file of the Unity Licensing Server is required when configuring and running the module. It can be downloaded from the `Organizations` section in the [Unity ID portal](https://id.unity.com/). For simplicity and cost optimization, this module requires the Linux version of the executable.
+
+The deployment uses an [Amazon EC2 instance](https://aws.amazon.com/pm/ec2/) to run the Unity Licensing Server. An [Amazon Elastic Network Interface](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html) (ENI) is attached to the instance to provide a static private (and public if configured) IP to separate network settings from the compute instance, as the Unity Licensing Server binds the license to the machine's identity, which includes the following information:
+
+- The server's MAC address
```
danielw-aws, line 8: This is good to explain the technical limitations around why we are using a static ip

modules/unity/floating-license-server/README.md
danielw-aws, line 1: I dont see anywhere here where it talks about the pfx file needed for https or if there is a way to set the software up with that. May want to add some language around that.

danielw-aws: If we are doing https termination at the alb layer call that out but there may be some customers that want to bind directly to the license server. This may be a feature enhancement down the line.

modules/unity/floating-license-server/examples/simple/dns.tf
danielw-aws, line 1: How does this work with a pfx cert and eni?

modules/unity/floating-license-server/main.tf
```diff
@@ -0,0 +1,385 @@
...
+  security_group_id = aws_security_group.unity_license_server_sg[0].id
+  description       = "Allow all outbound traffic"
+  ip_protocol       = "-1"
+  from_port         = -1
+  to_port           = -1
+  cidr_ipv4         = "0.0.0.0/0"
+}
+
+#############################################################
+# Elastic Network Interface and Elasti IP (if needed)
```
danielw-aws: Typo

modules/unity/floating-license-server/main.tf
```diff
@@ -0,0 +1,385 @@
...
+
+resource "aws_iam_instance_profile" "ec2_profile" {
+  name = "${var.name}-ec2-profile"
+  role = aws_iam_role.ec2_access_role.name
+}
+
+# Data source to get the latest Ubuntu 24.04 LTS AMI from Canonical Ltd AWS account
+data "aws_ami" "ubuntu_latest" {
+  most_recent = true
+  owners      = ["099720109477"]
```
danielw-aws, line 229: call out that this is the official account for ubuntu?

danielw-aws (approved):
Module looks good to me to move forward. One thing I do think we need to address is how PFX cert works as this is native functionality in the software that we should support. This can probably be a forth coming pr or feature.

modules/unity/floating-license-server/examples/simple/dns.tf
erikyb: Right. Wanted to keep this a simple as possible but you are right, pfx cert should be a feature we support down the road.

danielw-aws (approved):
Looks good to me. Still wanna chat about the PFX file and how we should approach that.

modules/unity/floating-license-server/main.tf
erikyb: Nice catch. Fixed.

modules/unity/floating-license-server/main.tf
erikyb: Done.

modules/unity/floating-license-server/README.md
erikyb: Makes sense. Will add pfx cert support as future enhancement.

PR #757, opened 2025-10-14
fix(teamcity): service connect with external clusters and add arbitrary env var support

gabebatista:
**Issue number:** N/A

## Summary

This PR fixes critical bugs in the TeamCity module's Service Connect configuration when using external ECS clusters and adds support for custom environment variables in build agent configurations. It also increases ephemeral storage for build agents to accommodate larger build workloads.

  - Fix Service Connect namespace to work with external ECS clusters by making it always created
  - Use dynamic block for service_connect_defaults to conditionally set namespace
  - Add local variable for service_connect_namespace_arn to simplify references
  - Fix teamcity_cluster_id output to handle both created and external clusters
  - Add support for injecting custom environment variables into the build agents through build_farm_config
  - Increase ephemeral storage from 20GB to 50GB for build agents

### Changes

This PR includes bug fixes and enhancements to the TeamCity module:

**Bug Fixes:**
- Fixed Service Connect namespace creation when using external ECS clusters. Previously, the namespace resource had a `count` condition that prevented it from being created when `cluster_name` was provided, causing agent-to-server communication to fail.
- Fixed `service_connect_defaults` block to only be set when creating a new cluster (not when using an external one)
- Fixed `teamcity_cluster_id` output to correctly return either the created or external cluster ID

**Enhancements:**
- Added support for custom environment variables in `build_farm_config` via an optional `environment` map, making the module more flexible for different build agent configurations
- Increased ephemeral storage from default 20GB to 50GB for build agents to accommodate larger build workloads (e.g., Unity builds)
- Added `service_connect_namespace_arn` local variable to simplify namespace references throughout the module

### User experience

**Before:**
- Users could not use the TeamCity module with an external ECS cluster because Service Connect namespace wasn't created, causing agents to fail connecting to the TeamCity server
- Build agents had limited storage (20GB) which was insufficient for large builds
- Users could not pass custom environment variables to build agents, limiting flexibility

**After:**
- TeamCity module works correctly with both self-managed and external ECS clusters
- Build agents have 50GB of ephemeral storage, sufficient for Unity and other large builds
- Users can pass arbitrary environment variables to specific build agent configurations through the `environment` map in `build_farm_config`

## Checklist

* [x] Changes are documented

<details>

**No**, this is not a breaking change. All changes are backward compatible:
 - The `environment` field in `build_farm_config` is optional with a default of `{}`
 - Existing configurations will continue to work without modification
 - The Service Connect namespace changes fix bugs without altering the module's interface

</details>

## Acknowledgment

### DIFF
--- a/modules/teamcity/agents.tf
+++ b/modules/teamcity/agents.tf
@@ -15,6 +15,11 @@ resource "aws_ecs_task_definition" "teamcity_agent" {
   memory                   = each.value.memory
   execution_role_arn       = aws_iam_role.teamcity_agent_task_execution_role.arn
   task_role_arn            = aws_iam_role.teamcity_agent_default_role.arn
+
+  ephemeral_storage {
+    size_in_gib = each.value.ephemeral_storage_gib
+  }
+
   container_definitions = jsonencode([
     {
       name      = "teamcity-agent"
@@ -22,17 +27,24 @@ resource "aws_ecs_task_definition" "teamcity_agent" {
       cpu       = each.value.cpu
       memory    = each.value.memory
       essential = true
-      environment = [
-        {
-          name  = "SERVER_URL"
-          value = "http://teamcity-server.teamcity-namespace:8111"
-        },
-        {
-          name  = "AGENT_NAME"
-          value = "${each.key}-agent"
-        },
-
-      ]
+      environment = concat(
+        [
+          {
+            name  = "SERVER_URL"
+            value = "http://teamcity-server.teamcity-namespace:8111"
+          },
+          {
+            name  = "AGENT_NAME"
+            value = "${each.key}-agent"
+          }
+        ],
+        [
+          for k, v in each.value.environment : {
+            name  = k
+            value = v
+          }
+        ]
+      )
       logConfiguration = {
         logDriver = "awslogs"
         options = {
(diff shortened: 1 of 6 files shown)


modules/teamcity/variables.tf
```diff
@@ -236,8 +236,10 @@ variable "build_farm_config" {
     desired_count = number
     cpu           = number
     memory        = number
+    environment   = optional(map(string), {})
   }))
-  default = {}
+  default     = {}
+  description = "Build agent configurations. Each agent can have custom environment variables via the 'environment' map."
```
danielw-aws: Is there documentation on where I can find environment variables to pass?

modules/teamcity/agents.tf
```diff
@@ -15,24 +15,36 @@ resource "aws_ecs_task_definition" "teamcity_agent" {
   memory                   = each.value.memory
   execution_role_arn       = aws_iam_role.teamcity_agent_task_execution_role.arn
   task_role_arn            = aws_iam_role.teamcity_agent_default_role.arn
+
+  ephemeral_storage {
+    size_in_gib = 50 # Increase from default 20GB to 50GB for Unity builds
+  }
```
ghost, line 21: Probably should be a variable w a default... doesn't need to be specifically for "Unity build"

PR #762, opened 2025-10-15
fix(teamcity): service connect namespace for external clusters and add custom env vars

gabebatista:
**Issue number:** N/A

## Summary

This PR fixes a bug in Service Connect namespace configuration when using externally-managed ECS clusters and adds support for customizable build agent configuration.

**Bug Fix**: Previously, the Service Connect namespace ARN was always referenced from the module-created cluster, causing errors when users provided an external cluster via `cluster_name`. The namespace ARN
now correctly references the appropriate cluster (external or module-created).

**Enhancements**: Build agents now support custom environment variables for non-sensitive configuration and configurable ephemeral storage to accommodate different build workload requirements.

### Changes

This PR fixes a bug where the Service Connect namespace wasn't created when using an external ECS cluster, causing service discovery to fail. It also adds support for custom environment variables and
configurable ephemeral storage for build agents.

**Bug Fixes:**
- Service Connect namespace is now always created, regardless of whether an external cluster is provided
- The `service_connect_defaults` block is now conditionally applied only when the module creates its own cluster
- Fixed `teamcity_cluster_id` output to correctly return the cluster ID for both scenarios

**Enhancements:**
- Build agents can now receive custom environment variables via the `environment` map in `build_farm_config` for non-sensitive configuration like service endpoints and feature flags
- Ephemeral storage size is now configurable per agent via `ephemeral_storage_gib` parameter (defaults to 20 GiB)
- Added "Build Agent Configuration" section to README with security guidance

### User experience

**Before:**
- Using an external ECS cluster would result in Service Connect failures due to missing namespace
- Build agents had hardcoded 50 GiB ephemeral storage with no way to customize
- No way to pass configuration to build agents without rebuilding container images

**After:**
- Service Connect works correctly with both module-created and external ECS clusters
- Build agents can receive custom environment variables for configuration (e.g., license server URLs, API endpoints)
- Ephemeral storage size is configurable per agent type with sensible 20 GiB default
- README provides clear documentation on how to use custom environment variables safely

## Checklist

* [x] Changes are documented

<details>

No - all enhancements use optional parameters with backward-compatible defaults.

</details>

## Acknowledgment

### DIFF
--- a/modules/teamcity/README.md
+++ b/modules/teamcity/README.md
@@ -7,6 +7,47 @@ The TeamCity server relies on shared file system for persistent storage of confi
 ## Deployment Architecture
 ![TeamCity Module Architecture](./assets/media/diagrams/teamcity-server-architecture.png)
 
+## Build Agent Configuration
+
+### Custom Environment Variables
+
+Build agents can be configured with custom environment variables through the `build_farm_config` variable. This allows you to pass non-sensitive configuration data such as service endpoints, feature flags, or other runtime settings to your build agents.
+
+Each build agent configuration supports an optional `environment` map where you can define key-value pairs that will be injected as environment variables into the agent container:
+
+```hcl
+module "teamcity" {
+  source = "path/to/teamcity/module"
+
+  # ... other configuration ...
+
+  build_farm_config = {
+    "my-build-agent" = {
+      image         = "my-registry/my-build-agent:latest"
+      cpu           = 2048
+      memory        = 4096
+      desired_count = 2
+
+      # Custom environment variables for non-sensitive configuration
+      environment = {
+        UNITY_LICENSE_SERVER_URL = "http://unity-license-server:8080"
+        BUILD_CACHE_BUCKET       = "my-build-cache-bucket"
+        CUSTOM_API_ENDPOINT      = "https://api.example.com"
+        ENABLE_DEBUG_LOGGING     = "true"
+      }
+    }
+  }
+}
+```
+
+The module automatically injects the following default environment variables into all build agents:
+- `SERVER_URL`: The TeamCity server URL for agent registration
+- `AGENT_NAME`: The name of the agent instance
+
+Any custom environment variables you provide will be merged with these defaults, allowing you to extend the agent configuration without overriding essential settings.
+
+**Security Note**: Do not pass sensitive data such as passwords, API keys, or tokens through environment variables. Use AWS Secrets Manager or IAM roles for secure credential management.
+
 ## Examples
 
 For example configurations, please see the [examples](https://github.com/aws-games/cloud-game-development-toolkit/tree/main/modules/teamcity/examples).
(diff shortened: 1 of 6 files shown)


modules/teamcity/README.md
```diff
@@ -7,6 +7,47 @@ The TeamCity server relies on shared file system for persistent storage of confi
...
+}
+```
+
+The module automatically injects the following default environment variables into all build agents:
+- `SERVER_URL`: The TeamCity server URL for agent registration
+- `AGENT_NAME`: The name of the agent instance
+
+Any custom environment variables you provide will be merged with these defaults, allowing you to extend the agent configuration without overriding essential settings.
+
+**Security Note**: Do not pass sensitive data such as passwords, API keys, or tokens through environment variables. Use AWS Secrets Manager or IAM roles for secure credential management.
```
ghost, line 49: We could add this functionality easily: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/specifying-sensitive-data.html

modules/teamcity/variables.tf
```diff
@@ -232,12 +232,23 @@ variable "aurora_instance_count" {
...
   type = map(object({
-    image         = string
-    desired_count = number
-    cpu           = number
-    memory        = number
+    image                 = string
+    desired_count         = number
+    cpu                   = number
+    memory                = number
+    environment           = optional(map(string), {})
```
ghost, line 239: yeah - just add a "secrets" option maybe... Something to consider.

modules/teamcity/README.md
ayatanb: wait do we want to do this though cuz isn't mostly everything else defined in secrets manager or env variables

modules/teamcity/agents.tf
```diff
@@ -15,24 +15,36 @@ resource "aws_ecs_task_definition" "teamcity_agent" {
...
+            name  = "SERVER_URL"
+            value = "http://teamcity-server.teamcity-namespace:8111"
+          },
+          {
+            name  = "AGENT_NAME"
+            value = "${each.key}-agent"
+          }
+        ],
+        [
+          for k, v in each.value.environment : {
```
ayatanb, line 42: what does this do?

modules/teamcity/README.md
ghost: You can pass environment variables here, and we should create an issue allowing an arbitrary number of secrets to be passed as well. the task definition has fields for both.

modules/teamcity/variables.tf
```diff
@@ -232,12 +232,23 @@ variable "aurora_instance_count" {
...
   }))
-  default = {}
+  default     = {}
+  description = <<-EOT
+    Map of build agent configurations where each key is the agent name and the value defines:
+    - image: Container image for the build agent
+    - desired_count: Number of agent instances to run
+    - cpu: CPU units to allocate (1024 = 1 vCPU)
+    - memory: Memory in MiB to allocate
+    - environment: Optional map of custom environment variables for non-sensitive configuration
```
ayatanb, line 249: can someone explain why we need this optional environment mapping? I saw the "Enhancements: Build agents now support custom environment variables for non-sensitive configuration and configurable ephemeral storage to accommodate different build workload requirements" but I guess I don't understand what goes into different custom build workloads.

ayatanb (approved):
just asked some clarifying questions but looks like a good fix to something I overlooked!

modules/teamcity/agents.tf
ghost: it iterates over the map `each.value.environment` and writes each key value pair (k, v) into the container environment.

modules/teamcity/variables.tf
ghost: In Gabe's case it was for the Unity License Server I believe. He needed to be able to tell each agent where the ULS was, and an environment variable is the best way to do that.

Users will want to use custom containers for the agents, and those custom containers may be expecting environment variables. This gives users a way to specify those env variables when deploying.

ayatanb: is the mapping for k and v related to this?

PR #765, opened 2025-10-17
refactor: add additional iam policy support and fix coalesce error

novekm:
- Add `additional_policy_arns` in variables.tf
- Update references in locals.tf and iam.tf to use new field name
- Add documentation explaining field attaches IAM policy ARNs to instance roles
- Fix coalesce error for capacity_reservation_preference when both values are null

**Issue number:**
N/A

### Changes

This PR addresses two issues:

1. **Add `additional_policy_arns` field** - Adds a new field with clear naming that indicates it attaches IAM policy ARNs to the VDI instance role.

2. **Fix coalesce error for `capacity_reservation_preference`** - Resolves terraform plan failure when both `config.capacity_reservation_preference` and `var.capacity_reservation_preference` are null.

**Technical changes:**
- Add `additional_policy_arns` in variables.tf
- Update references in locals.tf and iam.tf to use new field name
- Add documentation explaining field attaches IAM policy ARNs to instance roles
- Fix coalesce error for capacity_reservation_preference when both values are null

### User experience

**Before:**
- Terraform plan would fail with coalesce error when capacity reservation preferences weren't set:
  ```
  Error: Error in function call
  Call to function "coalesce" failed: no non-null, non-empty-string arguments.
  ```

**After:**
- New `additional_policy_arns` field with clear naming that accurately describes its purpose
- Proper documentation with examples showing how to use AWS managed and custom policy ARNs
- Terraform plan works correctly when capacity reservation preferences are null
- Users can attach additional IAM policies like:
  ```hcl
  additional_policy_arns = [
    "arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess",
    "arn:aws:iam::123456789012:policy/MyCustomPolicy"
  ]
  ```

### User experience

## Checklist

* [x] Changes are documented

<details>
no
</details>

## Acknowledgment

### DIFF
--- a/modules/vdi/examples/private-connectivity/versions.tf
+++ b/modules/vdi/examples/private-connectivity/versions.tf
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">= 1.0"
+  required_version = ">= 1.13"
 
   required_providers {
     aws = {
--- a/modules/vdi/examples/public-connectivity/versions.tf
+++ b/modules/vdi/examples/public-connectivity/versions.tf
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">= 1.0"
+  required_version = ">= 1.13"
 
   required_providers {
     aws = {
--- a/modules/vdi/versions.tf
+++ b/modules/vdi/versions.tf
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">= 1.0"
+  required_version = ">= 1.13"
 
   required_providers {
     aws = {


modules/unity/accelerator/README.md
```diff
@@ -25,7 +25,7 @@ For example configurations, please see the [examples](https://github.com/aws-gam
 
 | Name | Version |
 |------|---------|
-| <a name="provider_aws"></a> [aws](#provider\_aws) | 6.15.0 |
+| <a name="provider_aws"></a> [aws](#provider\_aws) | 5.89.0 |
```
danielw-aws, line 28: this went down for version requirements?

modules/vdi/iam.tf
```diff
@@ -166,3 +166,21 @@ resource "aws_iam_role_policy_attachment" "vdi_cloudwatch_agent" {
   role       = aws_iam_role.vdi_instance_role[each.key].name
   policy_arn = "arn:aws:iam::aws:policy/CloudWatchAgentServerPolicy"
 }
+
+# Attach additional user-specified IAM policies to the default VDI role
+resource "aws_iam_role_policy_attachment" "additional_policies" {
```
danielw-aws, line 171: does this need to be aws_iam_role_policy_attachments_exclusive? do we have any clear patterns on when to use the two different attachment resources?

modules/unity/accelerator/README.md
danielw-aws: Do we have any standards on how to find this number?

modules/vdi/variables.tf
```diff
@@ -203,6 +208,9 @@ workstations = {
 
 # User assignment is now direct:
 # assigned_user = "alice"  # References users{} key directly in workstation
+
+additional_policy_arns: List of additional IAM policy ARNs to attach to the VDI instance role.
```
danielw-aws, line 212: Doesnt this need to be commented out?

modules/vdi/locals.tf
danielw-aws, line 1: Looks good

danielw-aws (dismissed):
Some minor comments on how we are defining minumum aws provider numbers and a question on the VDI inline example not having any comments. 

Why were there a number of readme updates in this that were not specific to the commit that got authored?

modules/vdi/variables.tf
novekm: Nah this is still part of the long heredoc multiline string since the variable is a long map(object) type

modules/vdi/iam.tf
novekm: I think we should keep it as the regular `aws_iam_role_policy_attachment`. The pattern I've been following is:

- `_exclusive` for module-owned policies (SSM, CloudWatch, etc.)

- Regular attachment for user-specified e.g. via `additional_policy_arns`

The _exclusive version would remove any policies users attach outside our module, which is conflicts with the ability of users to supply additional policies. Users should be able to attach stuff through our variable OR their own Terraform without conflicts. I think this is the pattern we should take, I'm just not sure if we allow users to supply their own policies for other modules right now

modules/unity/accelerator/README.md
novekm: right now dependabot does it I think every week or so when it submits version PRs, but yeah we should probably add a version check to pre-commit if it's not already there

modules/vdi/iam.tf
danielw-aws: We do so this should become the pattern. We need to stay away from exclusive then.

PR #766, opened 2025-10-20
fix: add g6 instance type support for driver installation in vdi module

bentcooke:
## Summary
The G6E EC2 instances provide better performance for high-end VDI due to the newer GPU.  This change enables the NVIDA driver to be installed.

### Changes
Including g6 instance type in the NVIDIA GRID driver install logic.

### User experience

Prior to this, if a G6e instance type was selected for VDI deployment, the NVIDIA drivers would not be installed and the user would need to do it manually.  With this change, G6e instances get the NVIDIA drivers installed during instance provisioning.

## Checklist

* [ ] Changes are documented

## Acknowledgment


assets/packer/virtual-workstations/shared/base_infrastructure.ps1
```diff
@@ -6,7 +6,7 @@ Write-Host "Starting shared VDI base infrastructure setup..."
 
 # Install NVIDIA GRID drivers if GPU instance
 $instanceType = (Invoke-RestMethod -Uri "http://169.254.169.254/latest/meta-data/instance-type" -TimeoutSec 10)
-if ($instanceType -match "^(g3|g4|g5|p2|p3|p4)") {
+if ($instanceType -match "^(g3|g4|g5|g6|p2|p3|p4)") {
```
gabebatista, line 9: Why not just use a wildcard here? `"^(g* | p*)"`

PR #779, opened 2025-11-04
fix(ci): optimize pre-commit workflow to prevent disk space errors

gabebatista:
**Issue number:** N/A

### Changes

Fixed the pre-commit GitHub Actions workflow failing with "no space left on device" errors during `terraform init` commands. The workflow runs `terraform init` across 25+ modules and examples, each
downloading large Terraform providers separately, causing disk exhaustion.

**Optimizations implemented:**

1. **Disk Space Cleanup**: Removes pre-installed tools not needed for pre-commit hooks (.NET SDK, Android SDK, Haskell compiler, CodeQL tools), freeing up ~14GB of disk space

2. **Terraform Plugin Cache**: Added `TF_PLUGIN_CACHE_DIR` environment variable to share downloaded providers across all `terraform init` calls, eliminating redundant downloads and saving ~9GB

3. **Dynamic Lock File Generation**: Generates `.terraform.lock.hcl` files with linux_amd64 checksums in CI before pre-commit runs, enabling plugin cache to work while keeping lock files out of the repository
 (preserving module flexibility for users)

4. **Lock File Caching**: Caches generated lock files based on a hash of all `versions.tf` files, dramatically reducing workflow time on subsequent runs when provider versions haven't changed

5. **Disk Usage Reporting**: Added `df -h` output before and after cleanup for visibility into disk space availability

### User experience

**Before:**
The pre-commit workflow failed with "no space left on device" errors when attempting to download and install Terraform providers. Workflows started with 22GB available and ran out of space at 36MB. PRs would
show failed checks, requiring manual investigation and re-runs.

**After:**
The pre-commit workflow completes successfully with 36GB available disk space after cleanup. Terraform providers are downloaded once and reused across all modules via the plugin cache. Lock files are
generated dynamically in CI based on the current `versions.tf` files and cached for reuse on subsequent runs.

**Performance improvements:**
- First run: ~5 minutes for lock file generation + normal pre-commit time
- Subsequent runs: ~1 second to restore cached lock files + normal pre-commit time
- Total space optimization: ~23GB freed/saved

Contributors see passing pre-commit checks without disk space failures, and the repository remains generic with users managing their own lock files locally.

## Checklist

* [x] Changes are documented

<details>

No, this is not a breaking change. This only modifies the GitHub Actions workflow environment setup and does not affect any code, modules, or how users interact with the repository. Lock files remain
gitignored, preserving the repository's flexibility as a reusable module library.

</details>

## Acknowledgment

### DIFF
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -5,13 +5,60 @@ permissions: read-all
 jobs:
   pre-commit:
     runs-on: ubuntu-latest
+    env:
+      # Share Terraform providers across all terraform init calls to reduce disk usage
+      TF_PLUGIN_CACHE_DIR: ${{ github.workspace }}/.terraform-plugin-cache
     steps:
+      - name: Free up disk space
+        # Remove pre-installed tools not needed for pre-commit hooks
+        # This prevents "no space left on device" errors during terraform init
+        run: |
+          echo "Disk space before cleanup:"
+          df -h
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /usr/local/lib/android
+          sudo rm -rf /opt/ghc
+          sudo rm -rf /opt/hostedtoolcache/CodeQL
+          echo "Disk space after cleanup:"
+          df -h
+
       - uses: actions/checkout@v5
       - uses: actions/setup-python@v6
         with:
           python-version: 3.12
       - uses: hashicorp/setup-terraform@v3
       - uses: terraform-linters/setup-tflint@v6
+
+      - name: Create Terraform plugin cache directory
+        run: mkdir -p $TF_PLUGIN_CACHE_DIR
+
+      - name: Generate hash of all versions.tf files
+        id: versions-hash
+        run: |
+          # Create a hash of all versions.tf files to use as cache key
+          HASH=$(find . -name "versions.tf" -not -path "*/.terraform/*" -not -path "*/.external_modules/*" -type f -exec sha256sum {} \; | sort | sha256sum | cut -d' ' -f1)
+          echo "hash=$HASH" >> $GITHUB_OUTPUT
+          echo "Versions hash: $HASH"
+
+      - name: Cache Terraform lock files
+        uses: actions/cache@v4
+        id: lock-files-cache
+        with:
+          path: "**/.terraform.lock.hcl"
+          key: terraform-lock-files-${{ steps.versions-hash.outputs.hash }}
+
+      - name: Generate lock files for CI environment
+        if: steps.lock-files-cache.outputs.cache-hit != 'true'
+        # Generate .terraform.lock.hcl files with linux_amd64 checksums for all modules
+        # This allows the plugin cache to work while keeping lock files out of the repo
+        run: |
+          echo "Generating lock files for all Terraform modules..."
+          find . -name "versions.tf" -not -path "*/.terraform/*" -not -path "*/.external_modules/*" | while read -r versions_file; do
+            dir=$(dirname "$versions_file")
+            echo "Processing: $dir"
+            (cd "$dir" && terraform init -backend=false && terraform providers lock -platform=linux_amd64) || echo "Failed to generate lock file for $dir"
+          done
+          echo "Lock file generation complete"
 
       - name: terraform-docs alias
         # This is necessary because terraform docs doesn't install super duper easily


novekm (approved):

.github/workflows/pre-commit.yaml
```diff
@@ -5,14 +5,61 @@ permissions: read-all
...
+        # This prevents "no space left on device" errors during terraform init
+        run: |
+          echo "Disk space before cleanup:"
+          df -h
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /usr/local/lib/android
+          sudo rm -rf /opt/ghc
+          sudo rm -rf /opt/hostedtoolcache/CodeQL
+          echo "Disk space after cleanup:"
+          df -h
```
ghost, line 23: are you noticing a difference here? I'd be shocked if GitHub isn't scrubbing runners between actions...

.github/workflows/pre-commit.yaml
```diff
@@ -5,14 +5,61 @@ permissions: read-all
...
+          HASH=$(find . -name "versions.tf" -not -path "*/.terraform/*" -not -path "*/.external_modules/*" -type f -exec sha256sum {} \; | sort | sha256sum | cut -d' ' -f1)
+          echo "hash=$HASH" >> $GITHUB_OUTPUT
+          echo "Versions hash: $HASH"
+
+      - name: Cache Terraform lock files
+        uses: actions/cache@v4
+        id: lock-files-cache
+        with:
+          path: "**/.terraform.lock.hcl"
+          key: terraform-lock-files-${{ steps.versions-hash.outputs.hash }}
```
ghost, line 48: Why are we doing this? Does the cache expire between runs of the pre-commit action? Is this just to prevent each Terraform step from re-initializing?

.github/workflows/pre-commit.yaml
gabebatista: The proof is in the pudding. I had the workflow output disk metrics before and after clean up. Check out the results:

``` bash
Disk space before cleanup:
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        72G   51G   22G  70% /
tmpfs           7.9G   84K  7.9G   1% /dev/shm
tmpfs           3.2G  1.1M  3.2G   1% /run
tmpfs           5.0M     0  5.0M   0% /run/lock
/dev/sdb16      881M   62M  758M   8% /boot
/dev/sdb15      105M  6.2M   99M   6% /boot/efi
/dev/sda1        74G  4.1G   66G   6% /mnt
tmpfs           1.6G   12K  1.6G   1% /run/user/1001
Disk space after cleanup:
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        72G   36G   36G  50% /
tmpfs           7.9G   84K  7.9G   1% /dev/shm
tmpfs           3.2G  1.1M  3.2G   1% /run
tmpfs           5.0M     0  5.0M   0% /run/lock
/dev/sdb16      881M   62M  758M   8% /boot
/dev/sdb15      105M  6.2M   99M   6% /boot/efi
/dev/sda1        74G  4.1G   66G   6% /mnt
tmpfs           1.6G   12K  1.6G   1% /run/user/1001
```

.github/workflows/pre-commit.yaml
gabebatista: The cache does not expire between runs, even across different PRs. The cache will only expire if it's not used for 7 days.

The reason this is set up is because the workflow was failing with "no space left on device" errors. Each of our 25+ modules was downloading Terraform providers independently, which was eating up like 9GB+ of
 disk space. To fix this, we implemented a Terraform plugin cache (TF_PLUGIN_CACHE_DIR) that lets all modules share the same provider downloads instead of each one downloading its own copy.

The catch is that the plugin cache needs consistent lock files with matching checksums across all modules. Without lock files, each terraform init generates fresh checksums that don't match the cached
providers, causing validation to fail. We can't just commit lock files to the repo though, since this is a reusable module library - that would force specific provider versions on users and kill the
flexibility they need.

So we generate lock files dynamically in CI before pre-commit runs. The issue is that generating lock files for all 25+ modules adds like 5-10 min to every workflow run. With the caching, we create a hash of
ALL versions.tf files in the repo (not individual files), and if any versions.tf changes anywhere, the hash changes and we regenerate all the lock files. But if none of the versions.tf files changed, we get a
 cache hit and restore all the lock files in about 1 sec instead.

This brings those 5-10 min down to 1 sec when cache hits, which should be quite often since most PRs will only change module code without touching the provider versions in versions.tf files.

PR #800, opened 2025-12-11
bug: Adding debug support to VDI module for conditional deletion protection

### LINKED ISSUE #788: Bug: VDI module deletion protection
### Expected Behaviour

As a safety guardrail, it is desirable to ensure deletion protection is enabled for all VDI resources. 

### Current Behaviour

Currently, even if manually enabling deletion protection manually on the instance (outside of Terraform) Terraform will disable this. This is expected. Terraform has no context into changes made outside of state, and deletion protection (controlled via the `disable_api_termination` attribute) defaults to `false` in the AWS provider.

### Possible Solution

Resolution for this would entail setting `disable_api_termination = true` in the module for **all** VDI instances. This would mean that running an option that could potentially destroy an instance would prevent the deletion, causing the terraform operation to fail as a guardrail. 

However, for CI/CD you will likely not want this, as it would cause deletion to fail for automation. We should also include a `debug` variable where if `debug = true` then `disable_api_termination = false`, allowing people to either disable deletion protection by:
- manually setting disable_api_termination = false
- manually setting debug = true

#### Examples
**Production Environment:**
```hcl
module "vdi" {
  source = "./vdi-module"
  debug  = false  # Termination protection ENABLED
  # other variables...
}
```

**CI/CD Environment :**
```hcl
module "vdi" {
  source = "./vdi-module"
  debug  = true   # Termination protection DISABLED
  # other variables...
}
```

### Steps to Reproduce
1. Manually enable deletion protection for the EC2 instance
2. Run `terraform plan`
3. Plan shows deletion protection will be disabled (as this is in terraform state)
4. Terraform apply makes that change

latest

hwkiem:
**Issue number:**
Closes #788 

## Summary
Added a debug flag that conditionally disables deletion protection via the API. Debug is disabled by default. This protects accidental deletion of VDI instances provisioned through the CGDTK.

_NOTE:_ Resources provisioned with Terraform and then subsequently modified through the console or otherwise are not protected by this. Running `terraform plan` or `terraform apply` will always attempt to modify your infrastructure to match the declarative configuration specified. If you suspect that your infrastructure has drifted you can run `terraform plan -refresh-only` to determine changes between your configuration and infrastructure. We recommend managing all infrastructure through Terraform to prevent the introduction of drift.

## Acknowledgment

### DIFF
--- a/.github/workflows/terraform-validation.yml
+++ b/.github/workflows/terraform-validation.yml
@@ -39,7 +39,7 @@ jobs:
           elif [ "${{ github.event_name }}" = "push" ] && [ "${{ github.ref }}" = "refs/heads/main" ]; then
             dirs=$(find modules samples -name "*.tf" -type f | xargs -I {} dirname {} | sort -u | jq -R -s -c 'split("\n")[:-1]')
           elif [ "${{ steps.filter.outputs.terraform }}" = "true" ]; then
-            dirs=$(echo "$CHANGED_FILES" | jq -r '.[]' | xargs -I {} dirname {} | sort -u | jq -R -s -c 'split("\n")[:-1]')
+            dirs=$(echo "$CHANGED_FILES" | jq -r '.[]' | xargs -I {} dirname {} | sort -u | while read dir; do [ -d "$dir" ] && echo "$dir"; done | jq -R -s -c 'split("\n")[:-1]')
           else
             dirs="[]"
           fi


hwkiem:
I also cleaned up broken test directory. Sorry @novekm :)

novekm (dismissed):
LGTM

jorisdon (dismissed):
LGTM! Thanks for adding this! :) Not sure about the inclusion of the removal of the broken tests here, but approving because @novekm who owns that approved as well.

gabebatista (dismissed):

PR #813, opened 2025-12-13
Unreal Horde Source Build Documentation with ARM64 Support

### LINKED ISSUE #743: Feature request: Documentation for building Unreal Engine Horde from source with server.json configuration
### Use case

The Cloud Game Development Toolkit needs to transition from Epic's pre-built Horde container images to building from source. Epic hasn't updated their pre-built image in over a year, and attempts to inject `server.json` configuration into running containers have proven difficult and hacky.

Building from source will allow users to:
- Access current Horde features and security updates
- Cleanly integrate `server.json` configuration during build time
- Eliminate hacky runtime configuration workarounds
- Follow proper infrastructure-as-code practices

### Solution/User Experience

Create comprehensive documentation that guides users through building Horde containers from Unreal Engine source code with integrated `server.json` configuration. Users will follow a standardized build process that produces up-to-date, properly configured containers ready for deployment.

**Expected User Experience:**
```bash
# Clone Unreal Engine source
git clone [unreal-engine-repo]
cd Engine/Source/Programs/Horde

# Configure server.json for deployment
cp templates/server.json.example server.json
# Edit server.json with deployment-specific settings

# Build container with integrated configuration
docker build -t horde:latest.
docker tag horde:latest my-registry/horde:v1.0.0
```

The documentation will provide step-by-step instructions covering prerequisites, build environment setup, server.json configuration options, container build process, and terraform integration. This approach eliminates runtime configuration hacks, ensures access to current Horde features, and maintains consistency with the toolkit's infrastructure-as-code principles. Users will have a clean, repeatable build process that produces properly configured containers without the need for post-deployment configuration injection.

## Simple server.json Example

```json
{
  "database": {
    "connectionString": "Server=localhost;Database=Horde;Integrated Security=true"
  },
  "auth": {
    "method": "Anonymous"
  },
  "storage": {
    "backends": [
      {
        "id": "default-s3",
        "type": "S3",
        "bucket": "my-horde-storage-bucket",
        "region": "us-east-1"
      }
    ]
  },
  "compute": {
    "clusters": [
      {
        "id": "default",
        "name": "Default Cluster"
      }
    ]
  },
  "logging": {
    "level": "Information"
  }
}
```

### Links

- [Unreal Engine Horde Documentation](https://dev.epicgames.com/documentation/en-us/unreal-engine/horde-in-unreal-engine)
- [`server.json` settings](https://dev.epicgames.com/documentation/en-us/unreal-engine/horde-settings-for-unreal-engine#serversettings)

cwwalb:
**Issue number:**
closes #743

### Changes

### User experience

## Checklist

* [ ] Changes are documented

<details>

</details>

## Acknowledgment

### DIFF
--- a/modules/unreal/horde/ecs.tf
+++ b/modules/unreal/horde/ecs.tf
@@ -31,20 +31,27 @@ resource "aws_ecs_task_definition" "unreal_horde_task_definition" {
   cpu                      = var.container_cpu
   memory                   = var.container_memory
 
+  runtime_platform {
+    operating_system_family = var.operating_system
+    cpu_architecture        = var.horde_server_architecture
+  }
+
   volume {
     name = "unreal-horde-config"
   }
 
   container_definitions = jsonencode(concat([
-    {
-      name  = var.container_name
-      image = var.image
-      repositoryCredentials = var.github_credentials_secret_arn != null ? {
-        "credentialsParameter" : var.github_credentials_secret_arn
-      } : null
+    merge({
+      name      = var.container_name
+      image     = var.image
       cpu       = var.container_cpu
       memory    = var.container_memory
       essential = true
+      }, var.github_credentials_secret_arn != null && !var.is_source_build ? {
+      repositoryCredentials = {
+        "credentialsParameter" : var.github_credentials_secret_arn
+      }
+      } : {}, {
       portMappings = [
         {
           containerPort = var.container_api_port
@@ -111,10 +118,10 @@ resource "aws_ecs_task_definition" "unreal_horde_task_definition" {
           condition     = "SUCCESS"
         }] : []
       )
-    },
+    }),
     {
       name                     = "unreal-horde-docdb-cert",
-      image                    = "public.ecr.aws/docker/library/bash:5.3",
+      image                    = "public.ecr.aws/docker/library/bash:latest",
       essential                = false
       command                  = ["wget", "https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem", "-P", "/app/config/"]
       readonly_root_filesystem = false
@@ -135,7 +142,7 @@ resource "aws_ecs_task_definition" "unreal_horde_task_definition" {
     }],
     local.need_p4_trust ? [{
       name      = "unreal-horde-p4-trust",
-      image     = "ubuntu:noble"
+      image     = "public.ecr.aws/ubuntu/ubuntu:noble"
       essential = false
       command = ["bash", "-exc", <<-EOF
         apt-get update
--- a/modules/unreal/horde/ecs.tf
+++ b/modules/unreal/horde/ecs.tf
@@ -121,7 +121,7 @@ resource "aws_ecs_task_definition" "unreal_horde_task_definition" {
     }),
     {
       name                     = "unreal-horde-docdb-cert",
-      image                    = "public.ecr.aws/docker/library/bash:latest",
+      image                    = "public.ecr.aws/docker/library/bash:5.3",
       essential                = false
       command                  = ["wget", "https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem", "-P", "/app/config/"]
       readonly_root_filesystem = false
--- a/modules/unreal/horde/variables.tf
+++ b/modules/unreal/horde/variables.tf
@@ -60,6 +60,32 @@ variable "image" {
   default     = "ghcr.io/epicgames/horde-server:latest-bundled"
 }
 
+variable "is_source_build" {
+  type        = bool
+  description = "Set this flag to true if you are using a custom built Horde Server image from source."
+  default     = false
+}
+
+variable "horde_server_architecture" {
+  type        = string
+  description = "The CPU architecture for Horde server container. Valid values: x86 or arm64"
+  default     = "X86_64"
+  validation {
+    condition     = contains(["X86_64", "ARM64"], var.horde_server_architecture)
+    error_message = "horde_server_architecture must be either 'X_86' or 'ARM64'"
+  }
+}
+
+variable "operating_system" {
+  type        = string
+  description = "The operating system for the Horde server container. Valid values: linux or windows"
+  default     = "WINDOWS_SERVER_2019_CORE"
+  validation {
+    condition     = contains(["LINUX", "WINDOWS_SERVER_2025_FULL", "WINDOWS_SERVER_2025_CORE", "WINDOWS_SERVER_2022_FULL", "WINDOWS_SERVER_2022_CORE", "WINDOWS_SERVER_2019_FULL", "WINDOWS_SERVER_2019_CORE"], var.operating_system)
+    error_message = "Operating_system must be either LINUX, WINDOWS_SERVER_2025_FULL, WINDOWS_SERVER_2025_CORE, WINDOWS_SERVER_2022_FULL, WINDOWS_SERVER_2022_CORE, WINDOWS_SERVER_2019_FULL, and WINDOWS_SERVER_2019_CORE."
+  }
+}
+
 variable "cluster_name" {
   type        = string
   description = "The name of the cluster to deploy the Unreal Horde into. Defaults to null and a cluster will be created."
--- a/modules/unreal/horde/variables.tf
+++ b/modules/unreal/horde/variables.tf
@@ -68,21 +68,21 @@ variable "is_source_build" {
 
 variable "horde_server_architecture" {
   type        = string
-  description = "The CPU architecture for Horde server container. Valid values: x86 or arm64"
+  description = "The CPU architecture for Horde server container. Valid values: X86_64 or ARM64"
   default     = "X86_64"
   validation {
     condition     = contains(["X86_64", "ARM64"], var.horde_server_architecture)
-    error_message = "horde_server_architecture must be either 'X_86' or 'ARM64'"
+    error_message = "horde_server_architecture must be either 'X86_64' or 'ARM64'"
   }
 }
 
 variable "operating_system" {
   type        = string
-  description = "The operating system for the Horde server container. Valid values: linux or windows"
-  default     = "WINDOWS_SERVER_2019_CORE"
+  description = "The operating system for the Horde server container. Valid values: LINUX, WINDOWS_SERVER_2019_CORE, etc."
+  default     = "LINUX"
   validation {
     condition     = contains(["LINUX", "WINDOWS_SERVER_2025_FULL", "WINDOWS_SERVER_2025_CORE", "WINDOWS_SERVER_2022_FULL", "WINDOWS_SERVER_2022_CORE", "WINDOWS_SERVER_2019_FULL", "WINDOWS_SERVER_2019_CORE"], var.operating_system)
-    error_message = "Operating_system must be either LINUX, WINDOWS_SERVER_2025_FULL, WINDOWS_SERVER_2025_CORE, WINDOWS_SERVER_2022_FULL, WINDOWS_SERVER_2022_CORE, WINDOWS_SERVER_2019_FULL, and WINDOWS_SERVER_2019_CORE."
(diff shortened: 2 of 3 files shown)


modules/unreal/horde/ecs.tf
```diff
@@ -31,20 +31,27 @@ resource "aws_ecs_task_definition" "unreal_horde_task_definition" {
...
+      name      = var.container_name
+      image     = var.image
       cpu       = var.container_cpu
       memory    = var.container_memory
       essential = true
+      }, var.github_credentials_secret_arn != null && !var.is_source_build ? {
+      repositoryCredentials = {
+        "credentialsParameter" : var.github_credentials_secret_arn
+      }
+      } : {}, {
```
hwkiem, line 54: Could move some of this logic to a `local`

modules/unreal/horde/ecs.tf
```diff
@@ -111,10 +118,10 @@ resource "aws_ecs_task_definition" "unreal_horde_task_definition" {
           condition     = "SUCCESS"
         }] : []
       )
-    },
+    }),
     {
       name                     = "unreal-horde-docdb-cert",
-      image                    = "public.ecr.aws/docker/library/bash:5.3",
+      image                    = "public.ecr.aws/docker/library/bash:latest",
```
hwkiem: It is probably safer to pin to a specific version of bash. :latest opens us up to upstream issues

hwkiem: I know we do this elsewhere... probably should open an issue and fix those too.

modules/unreal/horde/ecs.tf
```diff
@@ -135,7 +142,7 @@ resource "aws_ecs_task_definition" "unreal_horde_task_definition" {
     }],
     local.need_p4_trust ? [{
       name      = "unreal-horde-p4-trust",
-      image     = "ubuntu:noble"
+      image     = "public.ecr.aws/ubuntu/ubuntu:noble"
```
hwkiem, line 145: nice fix. I wonder if we should also pin this release to a particular version for the same reason as above.

modules/unreal/horde/source-build/README.md
hwkiem, line 1: Do you think we should put this in a nested directory like this? Or just update the Horde README with these instructions?

modules/unreal/horde/source-build/README.md
```diff
@@ -0,0 +1,122 @@
...
+    elif [ "$ARCH" = "arm64" ]; then \
+        rm -rf /app/out/runtimes/osx-x64 && \
+        rm -rf /app/out/runtimes/win-x64 && \
+        rm -rf /app/out/runtimes/win-x86 && \
+        rm -rf /app/out/runtimes/linux-x64* && \
+        rm -rf /app/out/runtimes/linux-x86 && \
+        rm -rf /app/out/runtimes/linux/native/libgrpc_csharp_ext.x64.so && \
+        rm -rf /app/out/runtimes/linux/native/libgrpc_csharp_ext.x86.so; \
+    fi
+```
```
hwkiem: What in the WORLD is going on here? Why are we removing the redis dependencies? I am confusion.

modules/unreal/horde/source-build/README.md
```diff
@@ -0,0 +1,122 @@
...
+2. Create a new build target to build the Hord Server with the Dashboard and push the image to your Amazon ECR repository. Be sure to replace the values for `<account-id>` and `<region>` with your own values.
+    ```xml
+    <Node Name="Build and Publish HordeServer to ECR" Requires="Build HordeServer;Build HordeDashboard">
+        <!-- Create a new image by combining server and dashboard image into one -->
+        <Docker-Build BaseDir="Engine/Source/Programs/Horde/HordeServer" Files="Dockerfile*" UseBuildKit="true" Tag="horde-server" DockerFile="Engine/Source/Programs/Horde/HordeServer/Dockerfile.dashboard" />
+
+        <!-- Publish the docker image to Amazon ECR -->
+        <Docker-Push Repository="<account-id>.dkr.ecr.<region>.amazonaws.com" Image="horde-server" TargetImage="horde/horde-server:$(Version)" AwsEcr="True" />
+    </Node>
+    ```
```
hwkiem: We can't use the pre-defined Horde buildgraph for this?

modules/unreal/horde/source-build/README.md
```diff
@@ -0,0 +1,122 @@
...
+## Update the Horde BuildGraph
+1. In an IDE, open the file `Engine/Build/BatchFiles/BuildGraph.xml`.
+2. Create a new build target to build the Hord Server with the Dashboard and push the image to your Amazon ECR repository. Be sure to replace the values for `<account-id>` and `<region>` with your own values.
+    ```xml
+    <Node Name="Build and Publish HordeServer to ECR" Requires="Build HordeServer;Build HordeDashboard">
+        <!-- Create a new image by combining server and dashboard image into one -->
+        <Docker-Build BaseDir="Engine/Source/Programs/Horde/HordeServer" Files="Dockerfile*" UseBuildKit="true" Tag="horde-server" DockerFile="Engine/Source/Programs/Horde/HordeServer/Dockerfile.dashboard" />
+
+        <!-- Publish the docker image to Amazon ECR -->
+        <Docker-Push Repository="<account-id>.dkr.ecr.<region>.amazonaws.com" Image="horde-server" TargetImage="horde/horde-server:$(Version)" AwsEcr="True" />
```
hwkiem: I do like this... Just wondering if we want to recommend the creation of a new graph. I wonder if you could raise a PR against the the engine itself with modifications / improvements to the Horde buildgraph.

modules/unreal/horde/source-build/README.md
```diff
@@ -0,0 +1,122 @@
...
+2. Under the module declaration `"unreal_engine_horde"` add the following line with the value of the image URI you noted earlier. Be sure to also set the value of `is_source_build` to `true` and set the value of `horde_server_architecture` to `ARM64` or `X_86` by passing the value with the `var.horde_server_architecture` variable as shown below:
+    ```python
+    module "unreal_engine_horde" {
+        ...
+        image = "<image-uri>"
+        is_source_build = true
+        horde_server_architecture = "<horde-server-architecture>"
+        ...
+    }
+    ```
```
hwkiem: We should manage this through variables on the example.

modules/unreal/horde/variables.tf
```diff
@@ -60,6 +60,32 @@ variable "image" {
...
+
+variable "horde_server_architecture" {
+  type        = string
+  description = "The CPU architecture for Horde server container. Valid values: x86 or arm64"
+  default     = "X86_64"
+  validation {
+    condition     = contains(["X86_64", "ARM64"], var.horde_server_architecture)
+    error_message = "horde_server_architecture must be either 'X_86' or 'ARM64'"
+  }
+}
```
hwkiem, line 77: Can we determine architecture via `data` block and a lookup against the image?

modules/unreal/horde/variables.tf
```diff
@@ -60,6 +60,32 @@ variable "image" {
   default     = "ghcr.io/epicgames/horde-server:latest-bundled"
 }
 
+variable "is_source_build" {
+  type        = bool
+  description = "Set this flag to true if you are using a custom built Horde Server image from source."
+  default     = false
+}
```
hwkiem, line 67: Why do we need to know if its a source build?

modules/unreal/horde/variables.tf
```diff
@@ -60,6 +60,32 @@ variable "image" {
...
+
+variable "operating_system" {
+  type        = string
+  description = "The operating system for the Horde server container. Valid values: linux or windows"
+  default     = "WINDOWS_SERVER_2019_CORE"
+  validation {
+    condition     = contains(["LINUX", "WINDOWS_SERVER_2025_FULL", "WINDOWS_SERVER_2025_CORE", "WINDOWS_SERVER_2022_FULL", "WINDOWS_SERVER_2022_CORE", "WINDOWS_SERVER_2019_FULL", "WINDOWS_SERVER_2019_CORE"], var.operating_system)
+    error_message = "Operating_system must be either LINUX, WINDOWS_SERVER_2025_FULL, WINDOWS_SERVER_2025_CORE, WINDOWS_SERVER_2022_FULL, WINDOWS_SERVER_2022_CORE, WINDOWS_SERVER_2019_FULL, and WINDOWS_SERVER_2019_CORE."
+  }
+}
```
hwkiem, line 87: Also this. I am not sure why we want to support a windows OS for Horde unless a customer explicitly asks for it. Open to your thoughts on this!

hwkiem (changes requested):
Great PR. A few minor changes and a couple discussions for the team.

modules/unreal/horde/source-build/README.md
cwwalb: This was the original code from the UE repository
```
# Remove native libs not used on Linux x86_64
RUN rm -rf /app/out/runtimes/osx* && \
    rm -rf /app/out/runtimes/win* && \
    rm -rf /app/out/runtimes/linux-arm* && \
    rm -rf /app/out/runtimes/linux/native/libgrpc_csharp_ext.x86.so
```

It seems to be removing libs that are not used for the Linux x86_64 because that is the architecture this is built for by default. I think this is to remove unnecessary libraries to remove bloat from the docker image. In order to follow this same principle we make this compatible with all compute architectures and OS by identifying the architecture and OS the build is for, then removing the other libraries.

modules/unreal/horde/variables.tf
cwwalb: I added that because in `ecs.tf` we have the current steps to use github credentials as part of the container definition to pull the public horde docker image. I didn't want to break the functionality so created the variable so the user can still use the pre-built docker image if they choose or go for the source build

modules/unreal/horde/variables.tf
cwwalb: That was something I wanted to ask. What are the most commonly used OS for Horde?

modules/unreal/horde/source-build/README.md
cwwalb: I agree also would make sense to do the same with the Horde Server Dockerfile since at the moment it only is designed to support AMD64 and I have the fixes to make it more adaptive to the different architectures which you see in the README

modules/unreal/horde/source-build/README.md
cwwalb: You can but would require more manual effort. I like that this builds and publishes in one go

modules/unreal/horde/source-build/README.md
hwkiem: Got it - thanks for clarifying.

A few thoughts:
1. You should raise a PR against the engine directly suggesting changes to the Horde dockerfile that make it truly multi-arch. Suggesting changes that customers need to make to their version of the engine is a little messier.
2. That being said, I think the work you've done to show how one can build Horde for ARM is valuable. It informs the PR, and you can and should include it in the documentation changes so that customers can take advantage of this before UE merges the changes upstream. Plus, you can link your PR/issue in the UE repo in the docs so people can upvote / endorse your changes.
3. The default of the toolkit should be support for x86 until the ARM support is first-party.

What do you think?

modules/unreal/horde/source-build/README.md
hwkiem: Again, maybe an upstream change should be suggested. Making modifications to source code outside of the toolkit is somewhat out of scope. I think including this path as an option / appendix documentation makes sense, but the default should be x86 since that's what the engine builds for by default, and using the existing build graph wherever possible. Disagree?

modules/unreal/horde/source-build/README.md
hwkiem: Exactly. Let's make default docs use default approach in the engine. Then create a change upstream to support multi-arch better, and link those changes in docs.

Want help with this? I'm happy to assist on the UE changes. Docs for now should default to X86 and then we include an optional section below that outlines all those QOL changes.

modules/unreal/horde/variables.tf
hwkiem: I think if they provide an image as a variable that accomplishes the same thing right?

1. Customer specifies "image" and we use it. Credentials are not required.
2. Customer does not specify image or arch. We use the latest version of horde published and require GHCR credentials?

modules/unreal/horde/variables.tf
hwkiem: We should really only support Linux distributions. I don't think this is required for containerized deployments...

PR #815, opened 2025-12-15
fix(p4): bump ebs volume mounting retry count

dislogical:
## Summary

We were seeing failures occasionally as the volumes weren't mounted in time.

### Changes

Bump EBS mount attempts from 3 to 100 to allow for slower mounts.

### User experience

Startup times may be slower in the failure case, but will be more resilient to slow boots.

## Checklist

* [X] Changes are documented

<details>
No.
</details>

## Acknowledgment

### DIFF
--- a/assets/packer/perforce/p4-server/p4_configure.sh
+++ b/assets/packer/perforce/p4-server/p4_configure.sh
@@ -541,10 +541,11 @@ perform_operations() {
 
 
 # Maximum number of attempts (added due to terraform not mounting EBS fast enough at instance boot)
-MAX_ATTEMPTS=3
+MAX_ATTEMPTS=5
 
 # Counter for attempts
 attempt=1
+delay=1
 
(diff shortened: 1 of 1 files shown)


assets/packer/perforce/p4-server/p4_configure.sh
```diff
@@ -541,7 +541,7 @@ perform_operations() {
 
 
 # Maximum number of attempts (added due to terraform not mounting EBS fast enough at instance boot)
-MAX_ATTEMPTS=3
+MAX_ATTEMPTS=100
```
hwkiem: Is 100 necessary? Do you know what the max number of attempts you saw?

dislogical: Probably not, we were seeing in the range of 4-5, but that wasn't until after I had set it this high. I'll take another pass and do a lower count but exponential backoff.

PR #852, opened 2026-01-06
feat(perforce/p4-code-review): Migrate from ECS to EC2

gabebatista:
**Issue number:** N/A

### Changes

This PR migrates the P4 Code Review (Helix Swarm) module from a containerized ECS/Fargate deployment to a native EC2 Auto Scaling Group deployment with persistent storage.

**Key changes:**

- **Packer Template**: New AMI builder for Ubuntu 24.04 LTS (officially supported by Perforce until 2029), replacing the unsupported Amazon Linux 2023
- **Terraform Module**: Complete infrastructure refactor from ECS to EC2 Auto Scaling with:
  - Persistent EBS volume support with automatic attachment on instance replacement
  - Self-healing infrastructure via Auto Scaling Group
  - Native DEB package installation instead of containers
  - SSM Session Manager access (replacing SSH)
  - Comprehensive user-data script for runtime configuration
- **P4 Server Integration**: 
  - Dual super user approach - always creates a `super` user for Swarm extension compatibility, plus optional custom super user
  - `unlimited_timeout` group for service accounts to prevent ticket expiration
  - Fixed Swarm extension configuration by establishing P4 trust and authentication before modifying extension settings
- **Documentation**: 
  - Enhanced Packer README with architecture details, runtime configuration, and troubleshooting guides
  - Updated p4-code-review module README to reflect EC2 architecture (removed ECS/Fargate/Docker references)
  - Updated variable descriptions in parent perforce module for EC2-based fields
- **Example**: Updated to remove deprecated ECS-related variables

### User experience

In the existing implementation, when the p4cr container is restarted jobs and queue data is lost due to the lack of Swarm-compatible persistent storage. The new implementation uses EC2 in place of ECS/Fargate, allowing for EBS volumes to handle persistent storage for `/opt/perforce/swarm/data`.

### Files Changed

| Area | Files | Description |
|------|-------|-------------|
| Packer | `assets/packer/perforce/p4-code-review/*` | New AMI builder for Ubuntu 24.04 with Swarm |
| Packer | `assets/packer/perforce/p4-server/p4_configure.sh` | Dual super user + unlimited_timeout group |
| Terraform | `modules/perforce/modules/p4-code-review/*` | EC2 ASG, EBS, launch template, user-data |
| Terraform | `modules/perforce/sg.tf` | Security group rules for P4 Server ↔ Swarm |
| Terraform | `modules/perforce/variables.tf` | Updated p4_code_review_config descriptions |
| Example | `modules/perforce/examples/create-resources-complete/main.tf` | Removed ECS variables |

### Prerequisites

1. AWS account with appropriate permissions
2. Packer installed (>= 1.8.0)
3. Terraform installed (>= 1.0)
4. A Route53 hosted zone for DNS records
5. An ACM certificate for HTTPS

### Step 1: Build the P4 Code Review AMI

```bash
cd assets/packer/perforce/p4-code-review
packer init p4_code_review_x86.pkr.hcl
packer build p4_code_review_x86.pkr.hcl
```

Note the AMI ID from the output (e.g., `ami-0abc123def456789`).

### Step 2: Build the P4 Server AMI (if not already available)

```bash
cd assets/packer/perforce/p4-server
packer init p4_al2023_x86.pkr.hcl
packer build p4_al2023_x86.pkr.hcl
```

### Step 3: Deploy the Infrastructure

Use the example configuration or create your own:

```bash
cd modules/perforce/examples/create-resources-complete

# Create terraform.tfvars with your values
cat > terraform.tfvars <<TFVARS
project_prefix = "test"
certificate_arn = "arn:aws:acm:us-east-1:ACCOUNT:certificate/CERT-ID"
route53_public_hosted_zone_id = "Z0123456789ABCDEFGHIJ"
fully_qualified_domain_name = "perforce.example.com"
TFVARS

terraform init
terraform apply
```

### Step 4: Verify the Deployment

1. **P4 Server**: Connect using P4V or p4 CLI
   ```bash
   p4 -p ssl:perforce.example.com:1666 info
   ```

2. **P4 Code Review (Swarm)**: 
   - Navigate to `https://review.perforce.example.com`
   - Log in with a Perforce user
   - Verify changes are visible and queue workers are running

3. **Verify Swarm Extension** (via SSM on P4 Server):
   ```bash
   # Check extension is installed
   p4 extension --list --type extensions
   
   # Verify Swarm-Secure is set correctly
   p4 extension --configure Perforce::helix-swarm -o | grep Swarm-Secure
   # Should show: Swarm-Secure: false
   ```

4. **Test Instance Recovery**:
   - Terminate the Swarm EC2 instance from AWS Console
   - Verify Auto Scaling Group launches a new instance
   - Confirm Swarm is accessible again with data intact (EBS volume reattaches)

### Expected Outcomes

- ✅ P4 Server accepts SSL connections on port 1666
- ✅ Swarm web UI loads and shows P4 changes
- ✅ Queue workers are running (check via Swarm API: `curl https://review.example.com/api/v10/queue/workers`)
- ✅ Swarm extension installed on P4 Server with `Swarm-Secure: false`
- ✅ Data persists across instance replacement

## Checklist

- [x] Changes are documented

<details>

**Yes, this is a breaking change.**

Users currently running P4 Code Review using the ECS-based module will need to:
1. Build new AMIs using the updated Packer template
2. Migrate their data from ECS containers to the new EBS-backed EC2 instances
3. Update their Terraform configurations to use the new EC2-based variables

Migration guide and detailed documentation have been included in the module README.
</details>

## Acknowledgment

### DIFF
--- a/modules/perforce/modules/p4-code-review/user-data.sh.tpl
+++ b/modules/perforce/modules/p4-code-review/user-data.sh.tpl
@@ -80,26 +80,49 @@ if [ "$ATTACHED_INSTANCE" == "$INSTANCE_ID" ]; then
 elif [ "$ATTACHED_INSTANCE" != "none" ] && [ "$ATTACHED_INSTANCE" != "null" ]; then
     log "Volume is attached to different instance $ATTACHED_INSTANCE - checking instance state"
 
-    # Check if the attached instance is terminated before force detaching
-    INSTANCE_STATE=$(aws ec2 describe-instances \
-        --region "$REGION" \
-        --instance-ids "$ATTACHED_INSTANCE" \
-        --query 'Reservations[0].Instances[0].State.Name' \
-        --output text 2>/dev/null || echo "unknown")
-
-    log "Previous instance $ATTACHED_INSTANCE state: $INSTANCE_STATE"
-
-    if [ "$INSTANCE_STATE" = "terminated" ] || [ "$INSTANCE_STATE" = "unknown" ]; then
-        log "Previous instance is terminated/unknown, safe to force detach"
-        aws ec2 detach-volume \
+    # Wait for attached instance to terminate (handles ASG replacement race condition)
+    # ASG launches new instance before terminating old one, so we may need to wait
+    MAX_WAIT=300
+    WAIT_TIME=0
+    while [ $WAIT_TIME -lt $MAX_WAIT ]; do
+        INSTANCE_STATE=$(aws ec2 describe-instances \
             --region "$REGION" \
-            --volume-id "$VOLUME_ID" \
-            --force 2>&1 | tee -a /tmp/swarm-setup.log || log "Warning: Force detach may have failed"
-    else
-        log "ERROR: Volume attached to running instance $ATTACHED_INSTANCE (state: $INSTANCE_STATE)"
+            --instance-ids "$ATTACHED_INSTANCE" \
+            --query 'Reservations[0].Instances[0].State.Name' \
+            --output text 2>/dev/null || echo "unknown")
+
+        log "Previous instance $ATTACHED_INSTANCE state: $INSTANCE_STATE"
+
+        if [ "$INSTANCE_STATE" = "terminated" ] || [ "$INSTANCE_STATE" = "unknown" ]; then
+            log "Previous instance is terminated/unknown, safe to force detach"
+            break
+        elif [ "$INSTANCE_STATE" = "shutting-down" ] || [ "$INSTANCE_STATE" = "stopping" ]; then
+            log "Previous instance is $INSTANCE_STATE, waiting for termination..."
+            sleep 10
+            WAIT_TIME=$((WAIT_TIME + 10))
+        elif [ "$INSTANCE_STATE" = "running" ]; then
+            # Instance still running - could be ASG replacement in progress
+            # Wait a bit to see if it starts terminating
+            log "Previous instance still running, waiting to see if ASG terminates it..."
+            sleep 10
+            WAIT_TIME=$((WAIT_TIME + 10))
+        else
+            log "ERROR: Unexpected instance state: $INSTANCE_STATE"
+            log "Cannot safely detach volume - manual intervention required"
+            exit 1
+        fi
+    done
+
+    if [ $WAIT_TIME -ge $MAX_WAIT ]; then
+        log "ERROR: Timed out waiting for previous instance to terminate after $${MAX_WAIT}s"
         log "Cannot safely detach volume - manual intervention required"
         exit 1
     fi
+
+    aws ec2 detach-volume \
+        --region "$REGION" \
+        --volume-id "$VOLUME_ID" \
+        --force 2>&1 | tee -a /tmp/swarm-setup.log || log "Warning: Force detach may have failed"
 
     # Wait for detachment with timeout
     log "Waiting up to 2 minutes for volume to become available..."


assets/packer/perforce/p4-code-review/p4-code-review-build.log
jorisdon, line 1: please remove this and .gitignore it

assets/packer/perforce/p4-code-review/README.md
hwkiem, line 1: I don't see changes to mkdocs.yaml, so this README does not get published to the pages site?

assets/packer/perforce/p4-code-review/README.md
```diff
@@ -0,0 +1,377 @@
...
+cd assets/packer/perforce/p4-code-review
+
+# 2. Initialize Packer (downloads required plugins)
+packer init p4_code_review_x86.pkr.hcl
+
+# 3. Validate the template
+packer validate p4_code_review_x86.pkr.hcl
+
+# 4. Build the AMI (takes ~10-15 minutes)
+packer build p4_code_review_x86.pkr.hcl
```
hwkiem, line 60: Does it build on ARM? Do we want to validate that doesn't work? If we could offer Graviton here that would be cost-effective.

assets/packer/perforce/p4-code-review/README.md
```diff
@@ -0,0 +1,377 @@
...
+### System Configuration
+
+- **AppArmor**: Ubuntu's security module (less restrictive by default for `/opt`)
+- **Services**: Apache2 and PHP-FPM enabled for automatic startup
+- **User**: `swarm` system user created with proper permissions
+- **Directories**: `/opt/perforce/swarm` prepared with correct ownership
+
+### What's NOT Configured Yet
+
+The following are configured at **runtime** when you launch an instance:
```
hwkiem: maybe we should call this at **deployment** rather than **runtime**

assets/packer/perforce/p4-code-review/README.md
```diff
@@ -0,0 +1,377 @@
...
+- EBS volume mounting for persistent data
+- Queue worker configuration (cron job and endpoint)
+- File permissions for worker processes
+- P4 Server extension installation (Swarm triggers)
+
+### Runtime Configuration Details
+
+When an EC2 instance launches, the user-data script performs the following steps:
+
+1. **EBS Volume Attachment**: Finds and attaches the persistent data volume by tags
```
hwkiem, line 116: Does user_data do this or do we handle attachment with Terraform?

assets/packer/perforce/p4-code-review/README.md
```diff
@@ -0,0 +1,377 @@
...
+2. **Filesystem Setup**: Creates ext4 filesystem (first launch) or mounts existing one
```
hwkiem, line 117: Should we use EFS or EBS for the persistent filesystem? If we need persistent filesystem only should we not just use container + EFS attachment?

assets/packer/perforce/p4-code-review/README.md
```diff
@@ -0,0 +1,377 @@
...
+   cat README.md
+   ```
+
+3. **Follow the deployment guide** in the module README, which covers:
+   - Creating AWS Secrets Manager secrets for P4 credentials
+   - Writing Terraform configuration
+   - Deploying the infrastructure
+   - Accessing the Swarm web console
+
+Or jump directly to: [P4 Code Review Module Documentation](../../../../modules/perforce/modules/p4-code-review/README.md)
```
hwkiem: I think this is excessive... instructing them to `cat README.md` is too descriptive. I think we just point them towards the module.

assets/packer/perforce/p4-code-review/p4_code_review_x86.pkr.hcl
```diff
@@ -0,0 +1,108 @@
...
+variable "ssh_interface" {
+  type = string
+  default = "public_ip"
+}
+
+variable "install_swarm_optional" {
+  type = bool
+  default = true
+  description = "Install helix-swarm-optional package (includes LibreOffice for document previews and ImageMagick for image previews). Adds ~500MB to AMI size."
+}
```
hwkiem, line 58: Why do we default this to true?

assets/packer/perforce/p4-code-review/p4_code_review_x86.pkr.hcl
```diff
@@ -0,0 +1,108 @@
...
+  source_ami_filter {
+    filters = {
+      name                = "ubuntu/images/hvm-ssd-gp3/ubuntu-noble-24.04-amd64-server-*"
+      architecture        = "x86_64"
+      root-device-type    = "ebs"
+      virtualization-type = "hvm"
+    }
+    most_recent = true
+    owners      = ["099720109477"]  # Canonical
+  }
```
hwkiem: I think we should use a data block to look up the latest. You're pinning this to a version of ubuntu. Do we want to do this? Also, do we want to offer versions that are not x86_64 and amd64 specific?

assets/packer/perforce/p4-code-review/swarm_instance_init.sh
```diff
@@ -0,0 +1,333 @@
...
+
+# Ensure data directory exists with proper ownership
+# Note: configure-swarm.sh will change these, we'll fix them again afterwards
+mkdir -p "$SWARM_DATA_PATH"
+chown -R swarm:www-data "$SWARM_DATA_PATH"
+chmod 775 "$SWARM_DATA_PATH"
+
+# Run the official P4 Code Review configuration script
+# This handles initial setup and P4 Server extension installation
+log_message "Running configure-swarm.sh with super user credentials"
```
hwkiem, line 122: so we have a script called `swarm_configure.sh` that calls the p4 provided `configure-swarm.sh`?

assets/packer/perforce/p4-code-review/swarm_instance_init.sh
hwkiem, line 1: general comment. I think we should shift away from writing our configuration scripts as shell scripts and instead use ansible playbooks and the ansible provisioner from Packer.

assets/packer/perforce/p4-code-review/swarm_configure.sh
```diff
@@ -0,0 +1,333 @@
...
+# The Perforce configure-swarm.sh script sets Swarm-Secure to true by default
+# We need to set it to false when using self-signed certificates
+log_message "Updating Swarm extension configuration for SSL settings"
+
+# Trust the P4 server SSL certificate and authenticate
+log_message "Establishing P4 connection for extension configuration"
+p4 -p "$P4D_PORT" trust -y 2>/dev/null || log_message "P4 trust already established or using non-SSL"
+echo "$P4D_SUPER_PASSWD" | p4 -p "$P4D_PORT" -u "$P4D_SUPER" login 2>/dev/null || {
+  log_message "WARNING: P4 login failed, extension configuration may not work"
+}
```
hwkiem: Bit confused here. I was under the impression that `configure-swarm.sh` was responsible for setting up the extension, but we're modifying it after the fact. Extension configuration exists on the p4 server (not on the swarm server), so I'd worry that modifying the extension config after wouldn't propagate to the server.

That being said, it looks like we aren't actually updating the config? We're just logging in and trusting the server.

assets/packer/perforce/p4-code-review/swarm_instance_init.sh
```diff
@@ -0,0 +1,333 @@
...
+    // Configure Redis connection for session storage and caching
+    if (!isset(\$config['redis'])) {
+      \$config['redis'] = array();
+    }
+    \$config['redis']['options'] = array(
+      'server' => array(
+        'host' => '$SWARM_REDIS',
+        'port' => $SWARM_REDIS_PORT,
+      ),
+    );
```
hwkiem, line 190: Didn't we just set this with the `configure-swarm.sh` call? We're now modifying it again?

assets/packer/perforce/p4-code-review/swarm_configure.sh
```diff
@@ -0,0 +1,333 @@
...
+      \$config['environment'] = array();
+    }
+    \$config['environment']['hostname'] = 'https://$SWARM_HOST';
+
+    // Write back the configuration
+    file_put_contents('$SWARM_CONFIG', '<?php' . PHP_EOL . 'return ' . var_export(\$config, true) . ';' . PHP_EOL);
+  " || {
+    log_message "ERROR: Failed to update config.php with Redis settings"
+    exit 1
+  }
```
hwkiem: I understand we need to manually set this in the config file since the script doesn't set all settings, but the block includes updates to the redis config. That was my main concern.

assets/packer/perforce/p4-code-review/swarm_configure.sh
```diff
@@ -0,0 +1,333 @@
...
+  # Configure SSO (Single Sign-On) if enabled
+  if [ "$ENABLE_SSO" = "true" ]; then
+    log_message "Enabling SSO in configuration"
+    php -r "
+      \$config = include '$SWARM_CONFIG';
+      \$config['saml']['sp']['entityId'] = 'https://$SWARM_HOST';
+      \$config['saml']['sp']['assertionConsumerService']['url'] = 'https://$SWARM_HOST/saml/acs';
+      \$config['saml']['sp']['singleLogoutService']['url'] = 'https://$SWARM_HOST/saml/sls';
+      file_put_contents('$SWARM_CONFIG', '<?php' . PHP_EOL . 'return ' . var_export(\$config, true) . ';' . PHP_EOL);
+    " || log_message "WARNING: Failed to add SSO configuration"
```
hwkiem: This block makes more sense. Have you tested this?

assets/packer/perforce/p4-code-review/swarm_setup.sh
```diff
@@ -0,0 +1,115 @@
...
+./aws/install
+rm -rf aws awscliv2.zip
+cd -
+
+# Add Perforce repository
+log_message "Adding Perforce repository"
+wget -qO - https://package.perforce.com/perforce.pubkey | gpg --dearmor | tee /usr/share/keyrings/perforce-archive-keyring.gpg > /dev/null
+echo "deb [signed-by=/usr/share/keyrings/perforce-archive-keyring.gpg] http://package.perforce.com/apt/ubuntu jammy release" | tee /etc/apt/sources.list.d/perforce.list
+
+# Note: Using jammy (22.04) repo as noble (24.04) may not be available yet
```
hwkiem: Going to need some elaboration on this. Why? Have you validated it?

modules/perforce/README.md
hwkiem, line 1: Why did we remove the README with the youtube video explaining how to use the module? Was this a previously discussed thing that I'm just forgetting?

modules/perforce/main.tf
```diff
@@ -146,18 +140,24 @@ module "p4_code_review" {
...
+  instance_subnet_id = var.p4_code_review_config.instance_subnet_id
 
   create_application_load_balancer = false
   internal                         = var.p4_code_review_config.internal
 
-  create_default_role = var.p4_code_review_config.create_default_role
-  custom_role         = var.p4_code_review_config.custom_role
+  # ElastiCache Redis
+  elasticache_node_count = var.p4_code_review_config.elasticache_node_count
+  elasticache_node_type  = var.p4_code_review_config.elasticache_node_type
```
hwkiem, line 160: Do we want to use serverless elasticache?

modules/perforce/modules/p4-code-review/ec2.tf
```diff
@@ -0,0 +1,163 @@
...
+  lifecycle {
+    prevent_destroy = false # Set to true in production to prevent accidental deletion
+  }
+}
+
+
+##########################################
+# Launch Template
+##########################################
+# Defines the EC2 instance configuration for ECS
```
hwkiem: This sounds wrong?

modules/perforce/modules/p4-code-review/iam.tf
```diff
@@ -10,151 +10,133 @@ resource "random_string" "p4_code_review" {
...
-      "secretsmanager:DescribeSecret",
-      "secretsmanager:BatchGetSecretValue"
-    ]
-    resources = [
-      var.super_user_username_secret_arn,
-      var.super_user_password_secret_arn,
-      var.p4_code_review_user_username_secret_arn,
-      var.p4_code_review_user_password_secret_arn,
+      "ec2:DescribeVolumes",
+      "ec2:DescribeInstances"
```
hwkiem, line 66: Still not sure why we aren't handling volume attachment with Terraform instead of via user_data

modules/perforce/modules/p4-code-review/user-data-native.sh.tpl
```diff
@@ -0,0 +1,287 @@
...
+$settings_php = [
+    'p4' => ['sso' => '${enable_sso == "true" ? "enabled" : "disabled"}'],
+];
+$user_config = include 'config.user.php';
+
+if($user_config == false) {
+    $user_config = [];
+}
+
+return array_replace_recursive($gen_config, $settings_php, $user_config);
```
hwkiem: This feels duplicate as well. I see us setting config in multiple places. Via user_data (template file here), and our config script which calls another config script (the perforce provided one). Its complicated, and I'm not sure its entirely necessary.

modules/perforce/modules/p4-code-review/user-data.sh.tpl
hwkiem, line 1: Whats the difference between this and user-data-native?

modules/perforce/modules/p4-server/variables.tf
```diff
@@ -246,7 +246,7 @@ variable "super_user_password_secret_arn" {
 
 variable "super_user_username_secret_arn" {
   type        = string
-  description = "If you would like to manage your own super user credentials through AWS Secrets Manager provide the ARN for the super user's username here. Otherwise, the default of 'perforce' will be used."
+  description = "If provided, creates an ADDITIONAL super user with this name alongside the required 'super' user for Swarm compatibility. Both users will have full super privileges."
```
hwkiem: Confusing description

modules/perforce/sg.tf
```diff
@@ -102,6 +102,24 @@ resource "aws_vpc_security_group_ingress_rule" "perforce_web_services_inbound_fr
...
+  security_group_id            = aws_security_group.perforce_web_services_alb[0].id
+  description                  = "Allows Perforce Web Services ALB to receive HTTP traffic from P4 Server for Swarm extension communication."
+  ip_protocol                  = "TCP"
+  from_port                    = 80
+  to_port                      = 80
+  referenced_security_group_id = module.p4_server[0].security_group_id
+  tags = merge(var.tags,
+    {
+      Name = "${var.project_prefix}-perforce-web-services-alb-p4-server-http"
+    }
```
hwkiem: Why? Shouldn't we just configure the extension to use HTTPS?

modules/perforce/sg.tf
```diff
@@ -183,10 +213,34 @@ resource "aws_vpc_security_group_ingress_rule" "p4_code_review_inbound_from_perf
...
+# Allows P4 Server to make HTTPS calls to Swarm for change validation
+resource "aws_vpc_security_group_egress_rule" "p4_server_outbound_to_p4_code_review" {
+  count                        = var.p4_code_review_config != null && var.p4_server_config != null && var.create_default_sgs ? 1 : 0
+  security_group_id            = module.p4_server[0].security_group_id
+  description                  = "Allows P4 Server to send HTTPS traffic to P4 Code Review for trigger validation."
+  from_port                    = 443
+  to_port                      = 443
+  ip_protocol                  = "TCP"
+  referenced_security_group_id = module.p4_code_review[0].application_security_group_id
+}
```
hwkiem: Why is the server communicating directly with the ec2 instance? The purpose of ALB + ASG target group for p4 code review is to allow consolidated networking path from P4 Server and users.

assets/packer/perforce/p4-code-review/README.md
gabebatista: Based on the feedback we got from Joris, Swarm does not support EFS due to some file locking issues. This is what pushed me to convert the module to EC2 in the first place.

assets/packer/perforce/p4-code-review/README.md
gabebatista: Added missing reference to doc and tested locally.

assets/packer/perforce/p4-code-review/README.md
gabebatista: Done

modules/perforce/modules/p4-code-review/ec2.tf
gabebatista: That was a little artifact left over from my initial attempt of just running the container on an EC2 capacity provider, but I quickly came to realize that was the lazy way out and just porting to EC2 would keep things simpler. Removed mention of ECS.

assets/packer/perforce/p4-code-review/README.md
gabebatista: The creation of the EBS volume is handled by tf but finding and attaching the volume from that point on is handled by the user_data script. This is done so that if the instance is replaced by the auto scaler the same EBS volume is attached to the new instance.

assets/packer/perforce/p4-code-review/README.md
gabebatista: Agreed, fixed.

assets/packer/perforce/p4-code-review/p4_code_review_x86.pkr.hcl
gabebatista: Those optional packages enable document and image previews in P4CR so I figured it should default to the "full" experience and users could disable if they needed to for some reason

assets/packer/perforce/p4-code-review/p4_code_review_x86.pkr.hcl
gabebatista: I agree that pinning to a version isn't great, I am moving this to a data block  and allowing users to pass in their preferred version (defaulting to latest). This change also requires that the swarm script check whether the selected AMI supports PHP8 (required by swarm) and install it if necessary.

assets/packer/perforce/p4-code-review/swarm_instance_init.sh
gabebatista: Agreed, honestly didnt even think of that. Lets keep that to a follow up PR so that we at least have something that works on main

assets/packer/perforce/p4-code-review/swarm_instance_init.sh
gabebatista: `swarm_configure.sh` renamed to `swarm_instance_init.sh`

assets/packer/perforce/p4-code-review/swarm_configure.sh
gabebatista: We are updating config on the server on lines 155 - 157. The reason the configurations are done in this script and not the P4 provided one is because I wanted to completely avoid making any changes to the Perforce provided files. If you dont think that is necessary I can move the changes over.

assets/packer/perforce/p4-code-review/swarm_instance_init.sh
gabebatista: The ECS setup used Perforce's Docker image which has a custom entrypoint that reads SWARM_REDIS env vars. Native install uses configure-swarm.sh which has no Redis support, so we add it manually. I wasn't able to find where else this was set. Can you point me to it if you did?

assets/packer/perforce/p4-code-review/swarm_configure.sh
gabebatista: I don't quite follow the concern, can you elaborate?

assets/packer/perforce/p4-code-review/swarm_configure.sh
gabebatista: Now that you mention it... I have not. I only tried logging in directly. I will test this before re-submitting for review

assets/packer/perforce/p4-code-review/swarm_setup.sh
gabebatista: To be honest I'm not sure why this comment got added. I think when I was writing this the noble package wasn't yet available, but it is now. Changing reference to use Noble instead.

modules/perforce/README.md
gabebatista: Restored missing README

modules/perforce/modules/p4-code-review/iam.tf
gabebatista: Because doing so would not properly re-attach the EBS volume in case of an instance being replaced by the auto-scaler

modules/perforce/modules/p4-code-review/user-data.sh.tpl
gabebatista: The difference is one is unused. Removed unused script.

modules/perforce/modules/p4-code-review/user-data-native.sh.tpl
gabebatista: There should only be 2 scripts, ours and Perforce's. The reason I created a second script rather than edits Perforce's is that I did not want to make changes to what Perforce provides. If you disagree we can easily combine both scripts into one.

modules/perforce/modules/p4-server/variables.tf
gabebatista: Perforce requires that the extension be installed using a user name "super", but this variable would attempt to create the extension using whatever user we defined here. The change is that now we install everything with the "super" user, then create an additional admin user based on what you pass in to this variable.

modules/perforce/sg.tf
gabebatista: fixed

assets/packer/perforce/p4-code-review/README.md
gabebatista: From what I can tell, P4CR/Swarm does not support Arm64 (despite their other offerings supporting it)

modules/perforce/sg.tf
gabebatista: Fixed, now using HTTPS

assets/packer/perforce/p4-code-review/README.md
hwkiem: Why are we using an autoscaler? Isn't the server single-instance?

assets/packer/perforce/p4-code-review/swarm_instance_init.sh
hwkiem: Can you create an issue covering those changes?

assets/packer/perforce/p4-code-review/swarm_instance_init.sh
hwkiem: [Here's the docs.](https://help.perforce.com/helix-core/helix-swarm/swarm/2024.3/Content/Swarm/setup.post.html#setup.post.using_your_redis_server) You are correct -this is the only way to configure an external redis server. Probably something worth making configurable. Swarm runs its own swarm instance alongside the server - this config is just for external redis instances. Perhaps out of scope for this PR, but lets create another well-scoped issue for follow up.

assets/packer/perforce/p4-server/README.md
```diff
@@ -40,17 +40,46 @@ An instance that is provisioned with this AMI will not automatically deploy a P4
...
-   <fully qualified domain name of P4 Server> \
-   <URL for P4Auth>
+   --p4d_type p4d_master \
+   --hx_depots /dev/sdf \
+   --hx_metadata /dev/sdg \
+   --hx_logs /dev/sdh \
+   --username <AWS Secrets Manager secret ID for username> \
+   --password <AWS Secrets Manager secret ID for password> \
+   --fqdn perforce.example.com \
+   --auth https://auth.perforce.example.com
```
hwkiem, line 51: Is there a reason we're modifying the p4-server README as well?

modules/perforce/modules/p4-code-review/ec2.tf
```diff
@@ -0,0 +1,162 @@
...
+    }
+  )
+}
+
+
+##########################################
+# Auto Scaling Group
+##########################################
+# Single-instance ASG provides automatic instance replacement if it fails
+# Min=1, Max=1 ensures only one instance runs at a time (Swarm doesn't scale horizontally)
```
hwkiem, line 112: Ah - now I see it.

modules/perforce/modules/p4-server/main.tf
```diff
@@ -16,7 +16,7 @@ resource "awscc_secretsmanager_secret" "super_user_username" {
   count         = var.super_user_username_secret_arn == null ? 1 : 0
   name          = "${local.name_prefix}-SuperUserUsername"
   description   = "The username for the created P4 Server super user."
-  secret_string = "perforce"
+  secret_string = "super"
```
hwkiem: probably should also be parameterized and support external variable setting... or maybe not. Should we just highlight this? We do this kind of thing all over the place. It isn't really a problem since the password is generated serverside.

modules/perforce/main.tf
hwkiem: bump

assets/packer/perforce/p4-code-review/swarm_configure.sh
hwkiem: You clarified elsewhere. Redis updates don't get handled by the script.

assets/packer/perforce/p4-code-review/README.md
gabebatista: The auto scaler is there to health check the ec2 instance and replace it if necessary. The autoscaler has a min and max of 1 instance

assets/packer/perforce/p4-code-review/swarm_instance_init.sh
gabebatista: Done: https://github.com/aws-games/cloud-game-development-toolkit/issues/865

assets/packer/perforce/p4-code-review/swarm_instance_init.sh
gabebatista: done: https://github.com/aws-games/cloud-game-development-toolkit/issues/866

assets/packer/perforce/p4-server/README.md
gabebatista: While making some changes to the P4-server packer script I found that using positional arguments were getting unwieldy, so I changed it to named arguments. These README changes are explaining the new available flags/arguments and describing how the "create new user" behavior works

modules/perforce/main.tf
gabebatista: Do we want to use it.... That's up for debate as the additional cost may not really add much value, especially considering the data stored here can be recreated (meaning we are not super concerned about this crashing and being relaunched). With that said I do like the idea of having it as an option. At the moment you can sorta do this through the "external" option for redis, where users can add any redis cluster they wish.

assets/packer/perforce/p4-server/README.md
hwkiem: We were using named arguments for p4-server before though

modules/perforce/modules/p4-server/main.tf
gabebatista: rewrote user creation so that super is always created and treat as a service account and a separate "perforce" user is created for admin purposes.

assets/packer/perforce/p4-server/README.md
gabebatista: The README for the p4-server packer stuff says otherwise. If it was already using named arguments then this change just updates the README to say exactly that. The other updates are there to explain the user creation changes I added.

PR #950, opened 2026-07-01
fix(unreal/horde): Horde config and Windows build agent hardening

bentcooke:
## Summary
The Horde module could not reliably stand up a working build pipeline against a Perforce depot. This PR fixes two root causes and adds opt-in Windows agent build hardening.

### Changes

Config delivery (server.json / globals.json -> /app/Data) 
- ecs.tf: add an `unreal-horde-data` volume mounted at /app/Data on both the app container and the (repurposed) docdb-cert init container. The init container now renders /app/Data/server.json (always) and /app/Data/globals.json (only when `config_globals_json` is non-empty — writing an empty file would make Horde load a blank config) before the app starts.
- local.tf: render `server_json` carrying `configPath`, `forceConfigUpdateOnStartup`, `enableNewAgentsByDefault`, and the Perforce connection under `plugins.build.perforce[]` (id/serverAndPort/credentials). Remove the legacy `Horde__Perforce__0__ServerAndPort` env var and the two `Horde__Perforce__0__credentials__*` secrets (`horde_service_secrets` is now empty).
- variables.tf: add `p4_username` / `p4_password` (sensitive), `config_path`, and `config_globals_json`. The server's Perforce connection is now rendered into server.json from p4_username/p4_password, so add validations requiring both whenever `p4_port` is set (the p4_super_user_*_secret_arn variables no longer feed the server's own connection).

Windows agent build hardening (opt-in, backward compatible) 
- `agent_enable_long_paths` -> LongPathsEnabled=1 (fixes LNK1104/MAX_PATH on deep engine plugin response-file paths).
- `agent_working_dir` (e.g. "C:\H") -> short agent WorkingDir, written to a DURABLE agent.json under C:\ProgramData\Epic\Horde\Agent that also pins the server URL so it survives the 5.7+ agent self-upgrade (which blanks appsettings.json).
- `agent_uba_compute_ports` (e.g. "7000-7010") -> host Windows Firewall inbound rule for UBA distributed compile workers (the SG allowing it is not enough).
- `agent_uba_horde_pool` / `agent_uba_max_workers` -> UBT BuildConfiguration.xml enabling UBA-over-Horde, with <Horde> as a TOP-LEVEL element (not nested under <UnrealBuildAccelerator>).
- p4.exe is installed on Windows agents when `p4_port` is set (UE build steps shell out to p4.exe, absent on a stock AMI).

### User experience

Before: with a Perforce depot configured, the Horde server never loaded projects/streams and could not authenticate to Perforce; Windows agents produced failing from-source / UBA builds even against a correctly configured server.

After: the server's Perforce connection and config path are delivered in a schema the current Horde image honors, so projects/streams load and authentication succeeds. The new agent knobs all default to today's behavior, so existing deployments are unaffected unless they opt in.

## Checklist

- [ x] I have performed a self-review of this change
- [ x] Changes have been tested
- [ x] Changes are documented

## Acknowledgment

### DIFF
--- a/modules/unreal/horde/README.md
+++ b/modules/unreal/horde/README.md
@@ -18,6 +18,31 @@ Unreal Engine Horde is only available through the Epic Games Github organization
 
 For example configurations, please see the [examples](https://github.com/aws-games/cloud-game-development-toolkit/tree/main/modules/unreal/horde/examples).
 
+## Server Configuration Delivery
+
+The Horde server reads its startup settings from `/app/Data/server.json`. An init container (`unreal-horde-init`) renders this file before the server starts, so the settings the current Horde image honors (config path, Perforce connection) are delivered as configuration files rather than environment variables.
+
+### Perforce connection
+
+To connect Horde to a Perforce server, set `p4_port` and provide the credentials through AWS Secrets Manager:
+
+1. Create a Secrets Manager secret containing a JSON object with the Perforce username and password Horde should connect as:
+
+    ```json
+    {"username": "horde-service-user", "password": "example-password"}
+    ```
+
+2. Pass the secret's ARN to the module via `p4_credentials_secret_arn`.
+
+The credentials are injected into the init container at startup using ECS-native Secrets Manager integration (the task execution role is granted `secretsmanager:GetSecretValue` on the secret automatically). The init container substitutes them into `server.json` under `plugins.build.perforce` in-memory — they never appear in the ECS task definition, CloudTrail, container logs, or Terraform state.
+
+### Horde configuration (globals.json)
+
+The `config_path` variable controls the Horde server's `configPath` setting:
+
+- Set `config_path = "globals.json"` together with `config_globals_json` (a JSON string) to have the init container write your Horde configuration to `/app/Data/globals.json`.
+- Set `config_path` to a Perforce depot path (e.g. `"//UE/Main/Config/globals.json"`) to have Horde load its configuration from your depot instead.
+
 <!-- TODO -->
 <!-- ## Deployment Instructions -->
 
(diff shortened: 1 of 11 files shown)


awsbilal (dismissed):
huge

gabebatista (changes requested):
Thanks for this PR, @bentcooke — the root causes you've identified are spot-on. Horde's move away from env var config and the agent self-upgrade issue are real pain points, and the Windows hardening knobs are well thought out.

I have a few requests before we can merge:

### 1. Credentials should stay in Secrets Manager

The current approach renders `p4_username`/`p4_password` as plaintext variables directly into the init container's shell command. This means credentials are visible in the ECS console (task definition JSON) and CloudTrail to anyone with ECS read access.

The module previously used Secrets Manager ARNs for this — let's keep that pattern rather than regressing. Instead of the new plaintext variables, accept a single Secrets Manager secret ARN (containing `{"username": "...", "password": "..."}`) and have the init container fetch it at runtime:

```bash
CREDS=$(aws secretsmanager get-secret-value --secret-id "$SECRET_ARN" --query SecretString --output text)
```

Then inject the values into `server.json` at startup. The task execution role already has Secrets Manager access for the DocDB cert path, so permissions shouldn't be an issue.

This also cleans up the variable surface — the existing `p4_super_user_username_secret_arn` / `p4_super_user_password_secret_arn` variables are no longer wired to anything after this PR. Either repurpose them for the new approach or remove them and introduce a single `p4_credentials_secret_arn`. We want to keep the public interface clean so users aren't confused about which variables actually take effect.

### 2. Fix the failing test

The `unit_test_perforce_integration` test in `03_auth_methods.tftest.hcl` sets `p4_port` without the new required variables, triggering the validation error. Please update the test.

### 3. Rename the init container

`unreal-horde-docdb-cert` now handles config rendering in addition to the cert download. Please rename it to `unreal-horde-init` to reflect its actual purpose.

### 4. Update the README and examples

Please add the new variables and Perforce config behavior to the module's README, and update the relevant example to show how to pass the credentials secret ARN and `config_path`. This helps users understand the new config delivery model without having to read the source.

---

The Windows agent hardening section looks great — opt-in, backward compatible, and well-commented. Happy to approve once the above are addressed. Let me know if you have questions on any of these.

bentcooke:
@gabebatista Thanks for the review — all four points are addressed in the latest push.

  **1. Credentials in Secrets Manager** — Done. The plaintext `p4_username`/`p4_password` variables are gone, replaced by a single `p4_credentials_secret_arn`
  containing `{"username": "...", "password": "..."}`. The unwired `p4_super_user_*_secret_arn` variables are removed, so the public interface is clean.

  One deliberate deviation from the suggested implementation: rather than calling `aws secretsmanager get-secret-value` from the init container, I kept
  ECS-native secret injection (`valueFrom` with JSON-key syntax, e.g. `arn:...:username::`). Two reasons: the init image (`bash:5.3`) has no AWS CLI, and a
  runtime CLI call would authenticate via the *task* role — the execution-role permissions you referenced only apply to ECS-managed `valueFrom` injection. This
  way the module keeps its original Secrets Manager pattern and the existing execution-role policy path. `server.json` is rendered into the task definition
  with `__P4_USERNAME__`/`__P4_PASSWORD__` placeholders, and the init script substitutes the injected env values at startup with JSON-safe escaping.
  Credentials never appear in the task definition, CloudTrail, Terraform state, or container logs (the previous `cat server.json` is also removed, since it
  would have echoed the rendered credentials to CloudWatch).

  **2. Failing test** — `unit_test_perforce_integration` now passes `p4_credentials_secret_arn` and asserts the new pattern: both JSON-key secret references on
  the init container, and placeholders (not values) in the rendered `server.json`. Also added `unit_test_perforce_missing_credentials` to cover the validation
  error path. Full suite: 13/13 passing on Terraform 1.14.2.

  **3. Init container rename** — `unreal-horde-docdb-cert` → `unreal-horde-init`, including the `dependsOn` reference and log stream prefix.
  
  **4. README + examples** — Added a "Server Configuration Delivery" section to the module README covering the secret format, the injection flow, and the two
  `config_path` modes (rendered `globals.json` vs. depot path); regenerated the tf-docs inputs table. The complete example now shows `p4_port` +
  `p4_credentials_secret_arn` wiring and loads `config_globals_json` from a `globals.json` file.

gabebatista (approved):
Looks good, thank you for the contribution!

PR #973, opened 2026-08-12
fix(lore/edge-pod): strip deprecation mark from region before templat…

### LINKED ISSUE #971: Bug: Lore edge-pod module templatefile() fails with "inconsistent result" due to deprecation marks on data.aws_region.current.name
### Expected Behaviour

terraform apply should successfully create the edge-pod EC2 instances and render the userdata template without errors.

### Current Behaviour

terraform apply fails on module.lore_edge_1.aws_instance.edge (and edge_2) with:
  
  Error: Error in function call
  
    on ../cloud-game-development-toolkit/modules/unreal/lore/modules/edge-pod/main.tf line 188, in resource "aws_instance" "edge":
    188:   user_data_base64 = base64encode(templatefile("${path.module}/userdata.sh.tpl", {
  
  Call to function "templatefile" failed: function returned an inconsistent result,
  was: cty.StringVal("#!/bin/bash...").WithMarks(cty.NewValueMarks(marks.Sensitive, marks.deprecation<name is deprecated. Use region instead.>)),
  now: cty.StringVal("#!/bin/bash...").WithMarks(cty.NewValueMarks(marks.Sensitive, marks.deprecation<name is deprecated. Use region instead.>)).
  
  The rendered userdata content is correct — the error is purely caused by Terraform's mark consistency checker rejecting the deprecation mark that propagates
  from data.aws_region.current.name through templatefile().

### Possible Solution

Strip value marks before passing into templatefile(). The join("", [...]) pattern produces an unmarked copy of the string:
  
  # In modules/unreal/lore/modules/edge-pod/main.tf
  
  locals {
    ecr_region   = join("", [data.aws_region.current.name])
    ecr_registry = "${data.aws_caller_identity.current.account_id}.dkr.ecr.${local.ecr_region}.amazonaws.com"
  }
  
  Then reference local.ecr_region instead of data.aws_region.current.name in the templatefile() variables.

### Steps to Reproduce

1. Deploy the Lore module with two edge pods using AWS provider v5.x+ and Terraform 1.5+
  2. terraform init -upgrade && terraform apply
  3. Write tier deploys successfully, but both edge-pod instances fail at the templatefile() call

latest

awsbilal:
Fixes #971

**Issue number:971**

## Summary

AWS provider v5.x+ attaches a deprecation mark to data.aws_region.current.name. When this marked value is passed into templatefile(), Terraform's consistency checker sees the marks as making the 'was' and 'now' values differ, causing:

  Error: Error in function call
  Call to function "templatefile" failed: function returned an inconsistent result

Fix: use join("", [...]) to produce an unmarked copy of the region string before passing it into templatefile().

### User experience

before: Lore dont work

after: Lore work

## Checklist

- [x] Changes are documented

<details>

</details>

## Acknowledgment

### DIFF
--- a/modules/unreal/lore/modules/edge-pod/main.tf
+++ b/modules/unreal/lore/modules/edge-pod/main.tf
@@ -7,6 +7,7 @@ data "aws_ssm_parameter" "al2023_ami" {
 locals {
   instance_family = split(".", var.instance_type)[0]
   is_arm64        = contains(["c8gd", "c8g", "c7gd", "c7g", "c7gn", "m7g", "m7gd", "m8g", "r7g", "r7gd", "r8g", "t4g", "im4gn", "is4gen", "x2gd", "hpc7g"], local.instance_family)
+  ecr_region      = join("", [data.aws_region.current.name]) # strip deprecation mark to avoid templatefile() inconsistency error
   ecr_registry    = split("/", var.container_image)[0]
   hmac_key        = var.hmac_key != null ? var.hmac_key : random_id.hmac[0].hex
 }
@@ -186,7 +187,7 @@ resource "aws_instance" "edge" {
   vpc_security_group_ids = [aws_security_group.edge.id]
 
   user_data_base64 = base64encode(templatefile("${path.module}/userdata.sh.tpl", {
-    ecr_region      = data.aws_region.current.name
+    ecr_region      = local.ecr_region
     ecr_registry    = local.ecr_registry
     ca_cert_pem     = var.ca_certificate_pem
     container_image = var.container_image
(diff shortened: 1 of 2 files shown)


bentcooke (dismissed):
This works for me.  Thanks for the fix.

sambiggins-aws (dismissed):
This fix works, but note that `data.aws_region.current.name` is being [removed in provider v7.0](https://github.com/hashicorp/terraform-provider-aws/issues/42468), so the `join("", [...])` workaround will eventually need to be replaced anyway.

The upstream [`EpicGames/lore/contrib/aws`](https://github.com/EpicGames/lore/blob/main/contrib/aws/main.tf#L2) avoids this by accepting region as a `var.region` input variable instead of reading it from a data source. That pattern is immune to deprecation marks. Might be worth aligning with upstream in a follow-up.

danielw-aws (dismissed):
LGTM

gabebatista:
The deprecation is just changing the variable name from "name" to "region" any reason we cant do the full fix rather than just the workaround?

awsbilal:
switched from the join("", [...]) workaround to the renamed data.aws_region.current.region attribute, which should be a more "permanent" fix. Also bumped AWS Provider from 5.0 to 6.0

